PR #4, opened 2017-10-09
Add TLS support

pmwebster:
We ended up adding this for our use internally and thought it could be useful for others :)


jdolitsky:
Yes! Thank you so much!
Somebody was literally just asking for this a few hours ago... Merged

PR #5, opened 2017-10-09
Removing accidental binary inclusion

pmwebster:
didn't mean to include the build the first time around.


jdolitsky:
ah thanks. didnt even see that

PR #9, opened 2017-10-13
Add handler to upload multipart form data

cjauvin:
This PR implements what has been proposed and discussed in https://github.com/chartmuseum/chartmuseum/issues/7, namely, a new handler allowing `multipart/form-data` uploads on the `api/charts` route:

```bash
curl -F "chart=@mychart-0.1.0.tgz" http://localhost:8080/api/charts
curl -F "prov=@mychart-0.1.0.tgz.prov" http://localhost:8080/api/charts
curl -F "chart=@mychart-0.1.0.tgz" -F "prov=@mychart-0.1.0.tgz.prov" http://localhost:8080/api/charts
```

The added complexity is mainly due to the fact that the new route should be able to handle both chart and provenance files (at once, or separately), and provide atomicity: if a chart and a provenance file are provided for instance, the chart should not be stored if the provenance file is somehow faulty, and vice versa: both should succeed or fail, together. Also note that although it is not strictly necessary (because the filenames can be obtained from the form data), I have reused the functions `repo.ChartPackageFilenameFromContent` and `repo.ProvenanceFilenameFromContent` as a validation mechanism. 

I'd like to thank my colleague @davidovich for the help and support!

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Powered by some great Go technology:
 - [zap](https://github.com/uber-go/zap) - for logging
 
 ## Things that have been said in Helm land
->"Finally!!" 
+>"Finally!!"
 
 >"ChartMuseum is awesome"
 
@@ -69,6 +69,12 @@ curl --data-binary "@mychart-0.1.0.tgz" http://localhost:8080/api/charts
 If you've signed your package and generated a [provenance file](https://github.com/kubernetes/helm/blob/master/docs/provenance.md), upload it with:
 ```bash
 curl --data-binary "@mychart-0.1.0.tgz.prov" http://localhost:8080/api/prov
+```
+
+Both files can also be uploaded at once (or one at a time) on the `api/charts` route using the `multipart/form-data` format:
+
+```bash
+curl -F "chart=@mychart-0.1.0.tgz" -F "prov=@mychart-0.1.0.tgz.prov" http://localhost:8080/api/charts
 ```
 
 ## Installing Charts into Kubernetes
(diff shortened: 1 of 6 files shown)


jdolitsky:
@cjauvin this is great, thanks so much 👍 

My one nitpick, can we remove the CLI options for this? and just hardcode "chart" and "prov" as the form field names? Seems like the type of thing that should not be configurable

davidovich:
LGTM

@jdolitsky I understand that the form field names config option is a bit obscure, but with it, we are able to use chartmuseum as a perfect drop-in (albeit with a small proxy in for api mismatches) from our previous implementation. For uploading, we had a call in this form (which is, as a chart upload api, all over the place (in ci-cd, in user scripts and within automation tools)):

```
curl --fail -F "chartfile=@mychart-0.1.0.tgz" http://chartserver
```
Note the **chartfile** field.

Because this is a user facing ui, we thought that allowing configurable field names would lessen the integration burden for a similar implementations as ours. Also, as this is a very simple curl call, an implementer could want to vary the field name to convey more information from the call.

Alternatively, could we just introduce a silent environment variable for this use case (and drop it from the official CLI options)?

jdolitsky:
@davidovich thanks for the detailed explanation. Looks like this may help others as well. Merged!

jdolitsky:
@cjauvin This seems to have broken the acceptance tests:
```
Uploading test chart package "mychart-0.1.0.tgz"
POST http://localhost:8080/api/charts
HTTP STATUS: 500
HTTP CONTENT: {"error":"cannot process Content-Type: "}
Expected HTTP status code to be 201 but was 500.
--



```

jdolitsky:
I think this should fix
https://github.com/chartmuseum/chartmuseum/commit/86f09fc6a1af568697a7fc28932c369d0f2129a3

PR #12, opened 2017-10-18
Add Prometheus metrics

### LINKED ISSUE #11: Add Prometheus metrics
Hi Josh,

I have another request which stems from a requirement in the business context where we'll use ChartMuseum (but would also make sense in a more general context I believe): add Prometheus metrics, using https://github.com/prometheus/client_golang.

Things it would be nice to (optionally) measure:

* Number of requests, by type
* Duration of requests
* Number of charts served
* Time it takes to regenerate index
* Etc.

What do you think? Of course if you agree once more, I'll be happy to propose an implementation.

cjauvin:
Here is a solution that fixes #11. It does two things:

1. Use a middleware (https://github.com/zsais/go-gin-prometheus) to inject metrics at the request level (essentially covering almost all our monitoring needs in one fell swoop!)
2. Add two extra gauges: one giving the total number of distinct charts served at any moment, and the other the total number of chart _versions_, at any moment

A caveat with (1): `go-gin-prometheus` uses the name of the target handler as its counter label, e.g. 

```
chartmuseum_requests_total{code="409",handler="github.com/chartmuseum/chartmuseum/pkg/chartmuseum.(*Server).(github.com/chartmuseum/chartmuseum/pkg/chartmuseum.postRequestHandler)-fm",host="localhost:8080",method="POST"} 1
```

which is harder to read (and possibly less useful in the long run, if the code changes) than the API routes.. What do you think? If you agree, I could propose a change on the external project.

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -43,6 +43,7 @@ func cliHandler(c *cli.Context) {
 		Debug:                  c.Bool("debug"),
 		LogJSON:                c.Bool("log-json"),
 		EnableAPI:              !c.Bool("disable-api"),
+		EnableMetrics:          !c.Bool("disable-metrics"),
 		AllowOverwrite:         c.Bool("allow-overwrite"),
 		ChartURL:               c.String("chart-url"),
 		TlsCert:                c.String("tls-cert"),
@@ -143,6 +144,11 @@ var cliFlags = []cli.Flag{
 		Name:   "log-json",
 		Usage:  "output structured logs as json",
 		EnvVar: "LOG_JSON",
+	},
+	cli.BoolFlag{
+		Name:   "disable-metrics",
+		Usage:  "disable Prometheus metrics",
+		EnvVar: "DISABLE_METRICS",
 	},
 	cli.BoolFlag{
 		Name:   "disable-api",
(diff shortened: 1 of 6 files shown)


jdolitsky:
Thanks for this, seems extremely straightforward :+1:

Few questions:
- Can we change the CLI flag from `--disable-metrics` to `--enable-metrics`? I would imagine majority of people will not be using Prometheus
- I do not have a lot of background on Prometheus. How do the metrics get to Prometheus? We don't need an additonal flag like `--prometheus-host=<host>` ?
- If something fails on the metrics side, will it affect the service?

I also agree with your comment about metrics being hard to read. It would be nice if that showed `POST /api/charts` instead or in addition to the handler. Maybe we can start by opening an issue on their GitHub.

Thanks again!

cjauvin:
> Can we change the CLI flag from --disable-metrics to --enable-metrics? I would imagine majority of people will not be using Prometheus

No problem if that's what you want of course, but I am under the impression that Prometheus is relatively popular, especially in the K8s/Helm community, where it is a "native citizen".

> I do not have a lot of background on Prometheus. How do the metrics get to Prometheus? We don't need an additonal flag like --prometheus-host=<host> ?

Prometheus is using a "pull" architecture (https://prometheus.io/docs/introduction/faq/#why-do-you-pull-rather-than-push?) so no, it is not required.

If something fails on the metrics side, will it affect the service?

Not at all!

jdolitsky:
That's very interesting. Lets keep it as is then. Thank you!

PR #13, opened 2017-10-30
Add function to limit request counter cardinality

cjauvin:
Following the acceptance of [this PR](https://github.com/zsais/go-gin-prometheus/pull/17), this implements such an actual mapping function in the context of the ChartMuseum. To be clear, its effect will be such that, for example:

1. A call to `/charts/foo-1.2.3.tgz` will increment `chartmuseum_requests_total{url="/charts/:filename"}`, instead of `chartmuseum_requests_total{url="/charts/foo-1.2.3.tgz"}`
2. A call to `/api/charts/foo/1.2.3` will increment `chartmuseum_requests_total{url="/api/charts/:name/:version"}`, instead of `chartmuseum_requests_total{url="/api/charts/foo/1.2.3"}`

Reviewers: @jdolitsky, @mattfarina, @davidovich and @rodcloutier

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 43ffa53a3671201d1632271e22efa7b82b21febf1c93e3ab8c9ea91086df55ec
-updated: 2017-10-26T08:23:40.672037942-05:00
+hash: 141a7d6afe632c03c92ba5776d0a4870d8cfc9f1c2cf3feed6ad534f4a17810d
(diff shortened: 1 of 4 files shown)


davidovich:
LGTM.

I see this as a preliminary implementation that adds value (limiting prometheus metrics that will have the tendency to ballon with high cardinality). 

If we get to the point where a more fine grained or configurable behavior is wanted, we can add this at a later PR.

PR #15, opened 2017-11-06
Fix concurrency issue in presence of download errors

### LINKED ISSUE #14: Race condition observed with high chart count
Hi, 

We have stumbled upon a problem with our CM instance where suddenly only about half of our Google Storage backed charts were being served (without any error or hint in the debug logs whatsoever). Upon inspection, we have found this to be due to a race condition in [`server.addIndexObjectsAsync`](https://github.com/kubernetes-helm/chartmuseum/blob/master/pkg/chartmuseum/server.go#L299-L320), where the function-global `err` variable can be set to non-nil by a first goroutine, while being reset just after by another, effectively fighting to set the error condition. The effect of this is that storage download errors can easily be hidden away by the concurrency logic. The more charts there are, the more likely this is to happen also. It's easy to simulate this problem by adding this block:

```
if rand.Float64() < 0.9 {
    return new(helm_repo.ChartVersion), fmt.Errorf("simulated IO error")
}
```
in [`server.getObjectChartVersion`](https://github.com/kubernetes-helm/chartmuseum/blob/master/pkg/chartmuseum/server.go#L339).

With this simulated error code added, the initial index creation code should very likely fail (rerun until it does if you get "lucky"), while it doesn't, which is wrong.

We have a fix on the burner for this coming soon.

davidovich:
Use channel based flow control to fix concurrency when a download error occurs. Read chartVersion from channel to add to the index as they arrive.

Add context management to cancel any pending downloading goroutines.

Fixes #14 

Thanks @cjauvin!


pkg/chartmuseum/server.go
```diff
@@ -296,39 +297,47 @@ func (server *Server) addIndexObjectsAsync(index *repo.Index, objects []storage.
...
+		go func(o storage.Object) {
+			select {
+			case <-ctx.Done():
+				return
+			default:
 				chartVersion, err := server.getObjectChartVersion(o, true)
 				if err != nil {
 					err = server.checkInvalidChartPackageError(o, err, "added")
+				}
+				if err != nil {
```
jdolitsky: This condition has already been met above, lines 317 and 318 can be removed

davidovich: With the current state of the code, this is unfortunately not true, as [checkInvalidChartPackageError](https://github.com/davidovich/chartmuseum/blob/96c5ef6426dad427292b606c84004f0b12c6a036/pkg/chartmuseum/server.go#L359) can nil out the error (in the case where the package is malformed in the store).

jdolitsky: ah i see, thanks for correcting.

PR #19, opened 2017-11-16
Fix index response times increasing with many uploads

### LINKED ISSUE #18: Chartmuseum cannot hold a constant response time when latency to storage is high
This was discussed in the #chartmuseum slack channel.

On Google Cloud Storage, we are running chartmuseum with around 60 charts and their versions. This makes about 800 versioned charts. The index update is causing latency problems in presence of new uploads. The server cannot keep up, and response times keep getting higher and higher, caused by a threading "bubble effect". This happens when the latency to storage is very high. It can also happen if the index is big.

Here is what happens currently:

Someone posts a new chart to storage.

1) /index.yaml detects the change
2) chartmuseum wants to recalculate the diff
3) another /index.yaml comes in (while 1) is not fninished),
   goto 2)

You see the picture. Every other index (while the first one is being calculated) is in wait state to redo the update, causing accumulated waits.

This issue is to track an upcoming PR which fixes the problem.

davidovich:
Index regeneration before this PR was causing a concurrency pile-up when
many requests were coming in at a fast pace.  This was causing the time
to return /index to always increase. The behavior was observed on a high
latency connection to Google Cloud Storage, but would also occur with
other network stores.

With this PR, the regeneration time is constant and thus the serving of
index.yaml is also constant.

The implementation takes a general approach to first make initial
network requests (to satisfy the object list from storage) and to pile
up subsequent requests while the initial one is being completed. The
same algorithm is used to update the in-memory cached index.  Note that
these requests need not be protected by a lock because they are only
ever executing on their own. See server.getChartList() for the
implementaion of this idea.

A refactor was needed to separate the fetch from the diff calculation to
allow separate calling of the network heavy operations. While doing so,
we also removed redundant calls to storage file list update.

Also made small low-hanging fruit style optimisations to index
manipulations.

Added request ID to all requests for better debugging. This will be
visible with the --debug flag. This was indispensable to diagnose
complex concurrent processing.

To test the before and after state, we have added the use of the
locusti.io loadtesting engine. A simple README in the loadtesting/
directory shows how to install locust (with pipenv) and loadtest
chartmuseum. This will prove useful in the future.

Fixes #18

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ REVISION := $(shell git rev-parse --short HEAD;)
 
 HAS_GLIDE := $(shell command -v glide;)
 HAS_PIP := $(shell command -v pip;)
+HAS_PIPENV := $(shell command -v pipenv;)
 HAS_VENV := $(shell command -v virtualenv;)
 HAS_GOVIZ := $(shell command -v goviz;)
 HAS_DOT := $(shell command -v dot;)
@@ -57,6 +58,14 @@ test: setup-test-environment
 .PHONY: testcloud
 testcloud: export TEST_CLOUD_STORAGE=1
 testcloud: test
+
+.PHONY: startloadtest
+startloadtest:
+ifndef HAS_PIPENV
+	@sudo pip install pipenv
+endif
+	@cd loadtesting && pipenv install
+	@cd loadtesting && pipenv run locust --host http://localhost:8080
 
 .PHONY: covhtml
 covhtml:
--- a/Makefile
+++ b/Makefile
@@ -17,12 +17,23 @@ endif
 	@glide install --strip-vendor
 
 .PHONY: build
-build: export GOARCH=amd64
-build: export CGO_ENABLED=0
-build:
-	@GOOS=linux  go build -v -i --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
+build: build_linux build_mac build_windows
+
+build_windows: export GOARCH=amd64
+build_windows:
+	@GOOS=windows go build -v --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
+	    -o bin/windows/amd64/chartmuseum cmd/chartmuseum/main.go  # windows
+
+build_linux: export GOARCH=amd64
+build_linux: export CGO_ENABLED=0
+build_linux:
+	@GOOS=linux go build -v --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
 	    -o bin/linux/amd64/chartmuseum cmd/chartmuseum/main.go  # linux
-	@GOOS=darwin go build -v -i --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
+
+build_mac: export GOARCH=amd64
+build_mac: export CGO_ENABLED=0
+build_mac:
+	@GOOS=darwin go build -v --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
 	    -o bin/darwin/amd64/chartmuseum cmd/chartmuseum/main.go # mac osx
 
 .PHONY: clean
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ HAS_GOVIZ := $(shell command -v goviz;)
 HAS_DOT := $(shell command -v dot;)
 HAS_AWS := $(shell command -v aws;)
 
+CM_LOADTESTING_HOST ?= http://localhost:8080
+
 .PHONY: bootstrap
 bootstrap:
 ifndef HAS_GLIDE
@@ -65,7 +67,7 @@ ifndef HAS_PIPENV
 	@sudo pip install pipenv
 endif
 	@cd loadtesting && pipenv install
-	@cd loadtesting && pipenv run locust --host http://localhost:8080
+	@cd loadtesting && pipenv run locust --host $(CM_LOADTESTING_HOST)
 
 .PHONY: covhtml
 covhtml:
--- a/pkg/chartmuseum/server.go
+++ b/pkg/chartmuseum/server.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/kubernetes-helm/chartmuseum/pkg/repo"
@@ -27,20 +29,30 @@ type (
 	Router struct {
 		*gin.Engine
 	}
-
+	fetchedObjects struct {
+		objects []storage.Object
+		err     error
+	}
+	indexRegeneration struct {
+		index *repo.Index
+		err   error
+	}
 	// Server contains a Logger, Router, storage backend and object cache
 	Server struct {
-		Logger                 *Logger
-		Router                 *Router
-		RepositoryIndex        *repo.Index
-		StorageBackend         storage.Backend
-		StorageCache           []storage.Object
-		StorageCacheLock       *sync.Mutex
-		AllowOverwrite         bool
-		TlsCert                string
-		TlsKey                 string
-		ChartPostFormFieldName string
-		ProvPostFormFieldName  string
+		Logger                  *Logger
+		Router                  *Router
+		RepositoryIndex         *repo.Index
+		StorageBackend          storage.Backend
+		StorageCache            []storage.Object
+		regenerationLock        *sync.Mutex
+		fetchedObjectsLock      *sync.Mutex
+		fetchedObjectsChans     []chan fetchedObjects
+		regeneratedIndexesChans []chan indexRegeneration
+		AllowOverwrite          bool
+		TlsCert                 string
+		TlsKey                  string
+		ChartPostFormFieldName  string
+		ProvPostFormFieldName   string
 	}
 
 	// ServerOptions are options for constructing a Server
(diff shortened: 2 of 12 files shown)


davidovich:
@jdolitsky Would be nice to have your review on this, thank you. 😃

Makefile
```diff
@@ -17,12 +17,23 @@ endif
...
 .PHONY: build
-build: export GOARCH=amd64
-build: export CGO_ENABLED=0
-build:
-	@GOOS=linux  go build -v -i --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
+build: build_linux build_mac build_windows
+
+build_windows: export GOARCH=amd64
+build_windows:
+	@GOOS=windows go build -v --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
```
jdolitsky: we will want to add something to the README for how to download/install the windows binary from S3

pkg/chartmuseum/server.go
```diff
@@ -27,20 +29,30 @@ type (
...
 	}
-
+	fetchedObjects struct {
+		objects []storage.Object
+		err     error
+	}
+	indexRegeneration struct {
+		index *repo.Index
+		err   error
+	}
```
jdolitsky: lets order stuff starting with lowercase either above or below uppercase, just not in the middle

pkg/chartmuseum/server.go
```diff
@@ -27,20 +29,30 @@ type (
...
-		TlsCert                string
-		TlsKey                 string
-		ChartPostFormFieldName string
-		ProvPostFormFieldName  string
+		Logger                  *Logger
+		Router                  *Router
+		RepositoryIndex         *repo.Index
+		StorageBackend          storage.Backend
+		StorageCache            []storage.Object
+		regenerationLock        *sync.Mutex
```
jdolitsky: same here (lowercase)

pkg/chartmuseum/server.go
```diff
@@ -127,7 +139,8 @@ func NewServer(options ServerOptions) (*Server, error) {
 		RepositoryIndex:        repo.NewIndex(options.ChartURL),
 		StorageBackend:         options.StorageBackend,
 		StorageCache:           []storage.Object{},
-		StorageCacheLock:       &sync.Mutex{},
+		regenerationLock:		&sync.Mutex{},
```
jdolitsky: formatting issue. should run through gofmt

pkg/chartmuseum/server.go
```diff
@@ -181,22 +199,81 @@ func loggingMiddleware(logger *Logger) gin.HandlerFunc {
 	}
 }
 
-func (server *Server) syncRepositoryIndex() error {
-	_, diff, err := server.listObjectsGetDiff()
-	if err != nil {
-		return err
+// getChartList fetches from the server and accumulates concurrent requests to be fulfilled
```
jdolitsky: should be block-style comment

pkg/chartmuseum/server.go
```diff
@@ -207,94 +284,84 @@ func (server *Server) listObjectsGetDiff() ([]storage.Object, storage.ObjectSlic
...
-	}
-
-	index := &repo.Index{
-		IndexFile: server.RepositoryIndex.IndexFile,
-		Raw:       server.RepositoryIndex.Raw,
-		ChartURL:  server.RepositoryIndex.ChartURL,
-	}
+func (server *Server) regenerateRepositoryIndexWorker(diff storage.ObjectSliceDiff, storageObjects []storage.Object, reqID string) (*repo.Index, error) {
+	server.Logger.Debugf("(%s) Regenerating index.yaml", reqID)
+    index := &repo.Index{
```
jdolitsky: formatting

pkg/chartmuseum/server.go
```diff
@@ -207,94 +284,84 @@ func (server *Server) listObjectsGetDiff() ([]storage.Object, storage.ObjectSlic
...
 }
 
-func (server *Server) removeIndexObject(index *repo.Index, object storage.Object) error {
+func (server *Server) removeIndexObject(index *repo.Index, object storage.Object, reqID string) error {
 	chartVersion, err := server.getObjectChartVersion(object, false)
 	if err != nil {
 		return server.checkInvalidChartPackageError(object, err, "removed")
 	}
-	server.Logger.Debugw("Removing chart from index",
+	server.Logger.Debugw(fmt.Sprintf("(%s) Removing chart from index", reqID),
```
jdolitsky: requestID should be passed the Debugw similar to name and version below (in all instances)

pkg/repo/index_test.go
```diff
@@ -65,10 +65,13 @@ func (suite *IndexTestSuite) TestRemove() {
 		for i := 5; i < 10; i++ {
 			chartVersion := getChartVersion(name, i, now)
 			suite.Index.RemoveEntry(chartVersion)
+			suite.Empty(suite.Index.HasEntry(chartVersion))
```
jdolitsky: formatting

scripts/setup_test_environment.sh
```diff
@@ -14,7 +14,9 @@ export PATH="$PWD/testbin:$PATH"
 export HELM_HOME="$PWD/.helm"
 
 main() {
-    check_env_vars
+    if [[ $TEST_CLOUD_STORAGE == 1 ]]; then
```
jdolitsky: nice addition!

pkg/chartmuseum/server.go
davidovich: Yes, this is how I implemented it first. But it made it harder to link the different requests together. Having a prefix request id makes it easy to see at a glance (without scanning to the right), the ids are all aligned.

pkg/chartmuseum/server.go
davidovich: Yes, this is how I implemented it first. But it made it harder to link the different requests together. Having a prefix request id makes it easy to see at a glance (without scanning to the right), the ids are all aligned.

The result is this:

```
(455) Removing chart from index ...
```

instead of this: 

```
Removing chart from index  { "Request ID": 455, "name" : "", "version": 0.0.0 }
```

I'll do what you prefer.

pkg/chartmuseum/server.go
```diff
@@ -154,11 +168,15 @@ func (server *Server) Listen(port int) {
 }
 
 func loggingMiddleware(logger *Logger) gin.HandlerFunc {
+	var requestID int64
 	return func(c *gin.Context) {
+		reqID := strconv.FormatInt(atomic.AddInt64(&requestID, 1), 10)
```
jdolitsky: What do these end up looking like?  like `"Request ID": 455` ? Will this result in collisions?

pkg/chartmuseum/server.go
jdolitsky: Ok, let do both then. Its useful to be able to parse when using `--log-json`. Can we specify `"reqID": 455` vs `"Request ID": 455`?

Also we may want to send this back in a response header such as `X-ChartMuseum-ID` (we can do this later).

davidovich: ok, I'll do that.

pkg/chartmuseum/server.go
davidovich: Yes, this is the source of all requests, so the id's are unique. 1 to 2^64, I don't think there will be collisions (unless you get 100 req per second for 361 billion years :-)). Joking aside, no you won't get collisions because of the atomic increment for each request.

pkg/chartmuseum/server.go
davidovich: done.

pkg/chartmuseum/server.go
jdolitsky: Still, we probably want some sort of hash no? If the server is restarted, the first incoming request ID is "1", which will have been used for first request last time service started.

Also, if CM is run on multiple replicas, the "1" may be seen X times

jdolitsky:
Just got locust up and running btw, very cool :+1:
What values are you guys using for # of users and hatch rate?

jdolitsky:
FYI - just found this as well:
https://github.com/atarantini/ginrequestid

Looks like it gens a RequestID and puts it in the header too.

jdolitsky:
Also https://github.com/rs/xid

pkg/chartmuseum/server.go
davidovich: Yes, you are right, and I now see your point of view. I was not using the request-id for other purposes than in-process identification, so my joke was misplaced. How did you envision the utility of a globally unique request id (on the headers eventually) in chartmuseum ? From the debugging point of view, a uuid type of request-id would not be very practical in the current output in the logs (too long).

jdolitsky: I agree actually, which is why I suggested adding it as a k/v instead of log message prefix. The header would essentially be for tracking/debugging purposes, so you could match server-side logs to client requests.

Have you heard of jq? https://stedolan.github.io/jq/

If the logs are in JSON using `--log-json` option, you can pipe them to jq and format them however you want in the case that you are trying to debug request IDs visually

davidovich: Yes, jq is excellent. But I must admit that my itch was soothed once I got the incremental ids (for debugging concurrency). I think using jq would have slowed the iteration process for me at the moment. Would you mind if we opened an issue to implement the more global request-id as you want it ? I think it is not the core of what this PR is about.

davidovich:
> What values are you guys using for # of users and hatch rate?

I did a 30 users and 2 users per second. But this was only to have enough concurrency on my system. I didn't do a real upper limit stress test because our backend network is not able to hold the charge.

pkg/chartmuseum/server.go
jdolitsky: ^ yes, good idea

Issue #21, opened 2017-11-22
proposal: Add X-Request-Id header


jdolitsky:
With recent changes (#19) each request is assigned a numeric "RequestId" in order to easily group log lines tied to a request.

My suggestion is to take this one step further, and send the RequestId to the client in the`X-Request-Id` HTTP header for further debugging, correlating client-side requests to associated log lines.

I'd also like to propose using a UUID instead of integer to prevent collisions across replicas.

This gin middleware seems to do everything we need: https://github.com/atarantini/ginrequestid

In addition, we may want to discuss changing the format of log lines, i.e. should we prefix every line with this long UUID, or make it a k/v similar to `"method": "GET"`?

@cjauvin @davidovich thoughts?

davidovich:
I think this is a good idea. I was thinking about this and thought that we should use both approaches, as they serve different use cases. One is for local debugging (the numerical id) and the other is for remote debugging server to client interactions. 

The existing numerical implementation holds the Request ID name, but it should change names, so that this proposal can take its place. 

With this variant, we would keep the log prefix numerical, and also populate, as you suggest, the json k/v UUID request id. The numerical id should never bleed into the header as it is not unique across replicas, but they are unique in the process space where they are observed.

davidovich:
fixed by #23

PR #23, opened 2017-11-27
Add X-Request-Id header

jdolitsky:
Implementation of proposal discussed in #21.

The `X-Request-Id` header is added to every response, courtesy of [atarantini/ginrequestid](https://github.com/atarantini/ginrequestid) middleware.

Added new methods to the Logger type, suffixed with the letter "c", that take gin context as an argument, automatically adding request id to each log line associated with a request (as well as prefixing log message with local request count @davidovich).

Also, server.go was starting to grow, so I moved some stuff into new files logging.go and cache.go.

### DIFF
--- a/pkg/chartmuseum/cache.go
+++ b/pkg/chartmuseum/cache.go
@@ -22,7 +22,6 @@ import (
 	"github.com/kubernetes-helm/chartmuseum/pkg/repo"
 	"github.com/kubernetes-helm/chartmuseum/pkg/storage"
 
-	"github.com/gin-gonic/gin"
 	helm_repo "k8s.io/helm/pkg/repo"
 )
 
@@ -38,7 +37,7 @@ type (
 )
 
 // getChartList fetches from the server and accumulates concurrent requests to be fulfilled all at once.
-func (server *Server) getChartList(c *gin.Context) <-chan fetchedObjects {
+func (server *Server) getChartList(log loggingFn) <-chan fetchedObjects {
 	ch := make(chan fetchedObjects, 1)
 
 	server.fetchedObjectsLock.Lock()
@@ -48,7 +47,7 @@ func (server *Server) getChartList(c *gin.Context) <-chan fetchedObjects {
 		// this unlock is wanted, while fetching the list, allow other channeled requests to be added
 		server.fetchedObjectsLock.Unlock()
 
-		objects, err := server.fetchChartsInStorage(c)
+		objects, err := server.fetchChartsInStorage(log)
 
 		server.fetchedObjectsLock.Lock()
 		// flush every other consumer that also wanted the index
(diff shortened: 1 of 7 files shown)


pkg/chartmuseum/logging.go
```diff
@@ -0,0 +1,107 @@
...
+		config.Level = zap.NewAtomicLevelAt(zap.InfoLevel)
+	}
+	logger, err := config.Build()
+	if err != nil {
+		return new(Logger), err
+	}
+	defer logger.Sync()
+	return &Logger{logger.Sugar()}, nil
+}
+
```
davidovich:
```golang

type loggerFn func(msg string, keysAndValues ...interface{})

func MakeDebugLogger(c *gin.Context, server *Server) loggerFn  {
    return func(msg string, keysAndValues ...interface{})) {
        return server.Logger.Debugc(c, msg, keysAndValues ...)
    }
}
```

pkg/chartmuseum/handlers.go
```diff
@@ -30,7 +30,7 @@ type (
 )
 
 func (server *Server) getIndexFileRequestHandler(c *gin.Context) {
-	index, err := server.syncRepositoryIndex(c.MustGet("RequestID").(string))
+	index, err := server.syncRepositoryIndex(c)
```
davidovich: Here we could call factory function to create a logger:
``` golang
logger := MakeDebugLogger(c, server)

index, err := server.syncRepositoryIndex(logger)
```

pkg/chartmuseum/cache.go
```diff
@@ -0,0 +1,277 @@
...
+		err     error
+	}
+	indexRegeneration struct {
+		index *repo.Index
+		err   error
+	}
+)
+
+// getChartList fetches from the server and accumulates concurrent requests to be fulfilled all at once.
+func (server *Server) getChartList(c *gin.Context) <-chan fetchedObjects {
```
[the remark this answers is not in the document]
davidovich: The resulting signature would be like this:
```
func (server *Server) getChartList(logger loggerFn) <-chan fetchedObjects {
```

pkg/chartmuseum/logging.go
```diff
@@ -0,0 +1,107 @@
+package chartmuseum
```
davidovich: Should'nt we also add a logging package here to further separation ?

jdolitsky: Do you mean move this out of pkg/chartmuseum? I like the idea of keeping all server-related stuff in here, new file simply to make server.go less cluttered

pkg/chartmuseum/cache.go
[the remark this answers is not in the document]
jdolitsky: Yes, did not think of that. Good idea.

pkg/chartmuseum/logging.go
davidovich: I meant:
```golang
package logging
```

jdolitsky: Oh, I'm not sure that makes sense. If we were to do that, we would probably want to break it out into its own directory (pkg/<package_name>/logger.go).

For example, see package declarations in all files here, they all have `package kube`: https://github.com/kubernetes/helm/tree/master/pkg/kube

davidovich: nevermind

Issue #24, opened 2017-12-01
Proposal: Support for Azure Blob Storage


philliphoff:
I'd like to see Azure Blob Storage as an option for the storage backend.

PR #25, opened 2017-12-01
Add Azure Blob storage backend

philliphoff:
Adds a backend implementation based on Azure Blob Storage. Its use and implementation is based on that for Amazon S3; I'm not a Go developer so I've tried to match it as closely as possible. By all means provide feedback on syntax, style, error handling, etc.

I wasn't able to get the acceptance tests running; any pointers related to their proper setup would be helpful.

Associated with #24.

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -81,6 +81,8 @@ func backendFromContext(c *cli.Context) storage.Backend {
 		backend = amazonBackendFromContext(c)
 	case "google":
 		backend = googleBackendFromContext(c)
+	case "azure":
+		backend = azureBackendFromContext(c)
 	default:
 		crash("Unsupported storage backend: ", storageFlag)
 	}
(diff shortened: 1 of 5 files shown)


jdolitsky:
@philliphoff This is awesome! Thank you, thank you, thank you!

Need to sign up for Azure and try this out. Will provide feedback asap.

cmd/chartmuseum/main.go
```diff
@@ -231,6 +242,21 @@ var cliFlags = []cli.Flag{
...
+	},
+	cli.StringFlag{
+		Name:   "storage-azure-key",
+		Usage:  "key to store charts for --storage-azure-key",
+		EnvVar: "STORAGE_AZURE_KEY",
+	},
+	cli.StringFlag{
+		Name:   "storage-azure-container",
+		Usage:  "container to store charts for --storage-azure-container",
+		EnvVar: "STORAGE_AZURE_KEY",
```
davidovich: I believe this should be STORAGE_AZURE_CONTAINER

philliphoff: Correct, thanks!

philliphoff:
Apologies, but there's some red tape on my side that I need to wade through before I can allow this PR to be committed, so closing this until that's done (but keeping my fork/branch).

jdolitsky:
@philliphoff sounds good, sorry it’s taken so long on our end. Will use your branch to continue testing

PR #32, opened 2018-01-11
Add restricted IAM policy

### LINKED ISSUE #31: Add restricted IAM Policy for S3 Storage in documentation
We already talked about this in #2.
The goal is to get a restricted IAM Policy for S3 Storage.

qrevel:
Closes #31
Every API call is working with this policy.


jdolitsky:
excellent thanks!

PR #35, opened 2018-01-29
optionally allow anonymous read for GET operations when basic auth is…

rawlingsj:
… enabled resolves kubernetes-helm/chartmuseum#34

This is just a suggestion but it seems to work nicely, not sure about the flag name `--auth-anonymous-get` though.


jdolitsky:
@rawlingsj thanks! I will check this out later tonight

jdolitsky:
@rawlingsj made a few changes in routes.go before merging to reduce duplication. Kept the `--auth-anonymous-get` flag name, I think it make sense. Thanks again!

rawlingsj:
yay! many thanks @jdolitsky

Issue #38, opened 2018-02-08
s3 storage: add support for server side encryption


hoesler:
It would be nice if chartmuseum could support sse with s3.

jdolitsky:
Closed fixed by #39

PR #39, opened 2018-02-08
feature(amazon stoarge): add sse option

hoesler:
PR for #38


jdolitsky:
@hoesler thanks for this, will check out shortly!

PR #40, opened 2018-02-08
Upgrade to helm v2.8.0

SlickNik:
- Upgrade helm v2.8.0 (including the packages from the k8s.io/helm path)
- Had to pin a couple of dependencies to ensure that cloud.google.com/go/storage continues to work


jdolitsky:
👍

PR #41, opened 2018-02-09
Make Dockerfile multistage

liamawhite:
Use docker build multistage functionality to have self-contained docker builds. This makes docker the only dependency required for the build.

Open for suggestions as to which Go version to build from, I just happened to use 1.8.3 as it's the version we use internally.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,2 @@
-*
-!bin/**
+bin/**
+vendor/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,14 @@
 # Build Binary
 FROM golang:1.8.3 as gobuild
+RUN apt-get update && \
+    apt-get -y install ca-certificates
 WORKDIR /go/src/github.com/kubernetes-helm/chartmuseum
 COPY . .
 RUN make bootstrap
 RUN make build_linux
 
 # Build Image
-FROM alpine:3.6
-RUN apk add --no-cache ca-certificates
+FROM scratch
+COPY --from=gobuild /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=gobuild /go/src/github.com/kubernetes-helm/chartmuseum/bin/linux/amd64/chartmuseum /chartmuseum
 ENTRYPOINT ["/chartmuseum"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,12 @@
+# Build Binary
+FROM golang:1.8.3 as gobuild
+WORKDIR /go/src/github.com/kubernetes-helm/chartmuseum
+COPY . .
+RUN make bootstrap
+RUN make build_linux
+
+# Build Image
 FROM alpine:3.6
 RUN apk add --no-cache ca-certificates
-COPY bin/linux/amd64/chartmuseum /chartmuseum
+COPY --from=gobuild /go/src/github.com/kubernetes-helm/chartmuseum/bin/linux/amd64/chartmuseum /chartmuseum
 ENTRYPOINT ["/chartmuseum"]


davidovich:
I'm not sure about the added value this brings, could you elaborate? 

But I have another question. Since chartmuseum is a go only program, could we inherit from scratch ? Then I would see the value of combining your PR work and using scratch as a base image. In that case, we would (or not ?) probably have problems with ca-certificates, but that can be configured. 

For reference, Kelsey Hightower blogged about this: https://medium.com/@kelseyhightower/optimizing-docker-images-for-static-binaries-b5696e26eb07. The specific code for certificates is here: https://github.com/kelseyhightower/contributors/blob/master/main.go#L63-L67

What do ou think ?

liamawhite:
So the idea is that a build pipeline no longer requires Go to be installed in order to build chart museum. This is how our internal pipeline works.

I've just updated the PR with to a `FROM scratch` image and used the root certs from the first stage. I have verified manually:

```
curl http://localhost:8080/index.yaml

{"L":"DEBUG","T":"2018-02-09T20:20:35.173Z","M":"Anonymous GET enabled"}
{"L":"DEBUG","T":"2018-02-09T20:20:35.173Z","M":"Fetching chart list from storage"}
{"L":"DEBUG","T":"2018-02-09T20:20:35.803Z","M":"Regenerating index.yaml"}
{"L":"DEBUG","T":"2018-02-09T20:20:35.803Z","M":"Loading charts packages from storage (this could take awhile)","total":1}
{"L":"DEBUG","T":"2018-02-09T20:20:36.011Z","M":"Adding chart to index","name":"strongswan","version":"1.0.0"}
{"L":"DEBUG","T":"2018-02-09T20:20:36.012Z","M":"index.yaml regenerated"}
{"L":"INFO","T":"2018-02-09T20:20:36.012Z","M":"Starting ChartMuseum","port":8080}
{"L":"DEBUG","T":"2018-02-09T20:27:15.592Z","M":"[1] Incoming request: /index.yaml","reqID":"9f6b5730-7f47-4010-b57a-b34c5b3f430f"}
{"L":"DEBUG","T":"2018-02-09T20:27:15.592Z","M":"[1] Fetching chart list from storage","reqID":"9f6b5730-7f47-4010-b57a-b34c5b3f430f"}
{"L":"INFO","T":"2018-02-09T20:27:16.168Z","M":"[1] Request served","path":"/index.yaml","comment":"","latency":"575.467923ms","clientIP":"172.17.0.1","method":"GET","statusCode":200,"reqID":"9f6b5730-7f47-4010-b57a-b34c5b3f430f"}
```

jdolitsky:
This is cool, and I like the fact that somebody can download the repo and just run `docker build`, but there are a few issues as it relates to our CircleCI build. Take a look at circle.yml and Makefile. 

`make acceptance` uses the built binary to determine if chartmuseum works with Helm CLI. One thing we could do if we want to keep the multistage build is test using the Docker image vs the binary. Thoughts on this @davidovich ?

The other challenge is that `make release` will upload the mac, linux, and windows binaries to an S3 bucket. So we would end up running `make build_linux` twice, once during Docker build and once before release. If there is a way to get the binary to the workspace during Docker build then I'm all for it. What I dont want is for this to result in a longer build (currently ~ 3 minutes 15 seconds).

davidovich:
I haven't used multistage docker yet, but as I understand it, for the acceptance tests it would confine the binary to an image, pushing the dependency on docker presence there (in order to run the executable). In other words, making the executable in a docker image form makes every usage need docker.

This might not be a big issue, but I think there are enough little other things (as @jdolitsky mentioned for the publishing stage) to take into account and fix before warranting the move.

liamawhite:
Ok so I think I need to do the following things:
- `make acceptance` should also test the built image
- Try to work out how `make release` could reuse the binary built in the Dockerfile.

With regards to the latter, I'm not sure there is a way to achieve it other than having a Dockerfile.release which is essentially the current Dockerfile and the new Dockerfile. I'm not sure this is a better solution though.

Also, I noticed the CircleCI tests haven't run against this PR do they require you to authorise the tests?

jdolitsky:
@liamawhite sorry for the delay here.

You are correct, CircleCI tests should be running on PRs. This hasn't been setup yet due to security reasons (e.g. you could leak our AWS creds, etc.). We need something similar to Kubernetes CI which allows owners to say "/ok-to-test".

I'm also planning to move from CircleCI to Codefresh, so I am working with our solutions architect to completely redesign and improve the build strategy. This PR is coming at a bad time, unfortunately.

I'm going to close this for now. Thank you though, this will all be taken into account for the CI redesign.

PR #43, opened 2018-02-12
return 409 for chart already exists

### LINKED ISSUE #42: Chart already exists should not be a 500
500 indicates something went wrong at the server side.

I suggest chartmuseum should return a 403 Forbidden (as it's not overwriting) or vaguely possibly 400 Bad Request (as you should not try to overwrite a chart).

jdolitsky:
Fixes #42

cc @davidovich 

I want to deprecate the --data-binary option, but it should be done at a later point, such as when making breaking API changes.


davidovich:
LGTM

PR #44, opened 2018-02-14
Support for Microsoft Azure Blob Storage

### LINKED ISSUE #24: Proposal: Support for Azure Blob Storage

jdolitsky:
Fixes #24


davidovich:
LGTM 
Nice job

Issue #47, opened 2018-02-19
Azure backend reloads chart packages despite no changes


jdolitsky:
In an Azure blob storage container with a single, unmodified chart (chartmuseum-0.3.3.tgz), index.yaml is regenerated every time it is requested, despite no changes in storage.

Example output:
```
{"L":"DEBUG","T":"2018-02-19T11:27:25.561-0600","M":"Fetching chart list from storage"}
{"L":"DEBUG","T":"2018-02-19T11:27:25.852-0600","M":"Regenerating index.yaml"}
{"L":"DEBUG","T":"2018-02-19T11:27:25.852-0600","M":"Loading charts packages from storage (this could take awhile)","total":1}
{"L":"DEBUG","T":"2018-02-19T11:27:26.006-0600","M":"Adding chart to index","name":"chartmuseum","version":"0.3.3"}
{"L":"DEBUG","T":"2018-02-19T11:27:26.007-0600","M":"index.yaml regenerated"}
{"L":"INFO","T":"2018-02-19T11:27:26.007-0600","M":"Starting ChartMuseum","port":8080}
{"L":"DEBUG","T":"2018-02-19T11:27:28.977-0600","M":"[1] Incoming request: /index.yaml","reqID":"c6c84c97-2335-4619-9a48-049126602c7e"}
{"L":"DEBUG","T":"2018-02-19T11:27:28.977-0600","M":"[1] Fetching chart list from storage","reqID":"c6c84c97-2335-4619-9a48-049126602c7e"}
{"L":"DEBUG","T":"2018-02-19T11:27:29.068-0600","M":"[1] Regenerating index.yaml","reqID":"c6c84c97-2335-4619-9a48-049126602c7e"}
{"L":"DEBUG","T":"2018-02-19T11:27:29.188-0600","M":"[1] Updating chart in index","name":"chartmuseum","version":"0.3.3","reqID":"c6c84c97-2335-4619-9a48-049126602c7e"}
{"L":"DEBUG","T":"2018-02-19T11:27:29.188-0600","M":"[1] index.yaml regenerated","reqID":"c6c84c97-2335-4619-9a48-049126602c7e"}
{"L":"INFO","T":"2018-02-19T11:27:29.188-0600","M":"[1] Request served","path":"/index.yaml","comment":"","latency":"211.369ms","clientIP":"::1","method":"GET","statusCode":200,"reqID":"c6c84c97-2335-4619-9a48-049126602c7e"}
```

Issue #46, opened 2018-02-17
Change behaviour for root API route.


liamawhite:
Currently,`/` returns a 404. We are interested in changing this behaviour to return a redirect to `/index.yaml`.

During a discussion on slack with @jdolitsky. He suggested that maybe this route could return the version of chart museum that is running, but thinking through this later I thought that this is a potential security risk. Generally, you should not tell users the version of an application you are running because then they can cross reference that against potential exploits.

I am happy to do the legwork but the exact solution should probably be a result of some discussion.

jdolitsky:
@liamawhite was chatting with @davidovich, it may not be in best interest to support the redirect to /index.yaml in case we want to change the functionality of this route in the future.

Instead, how do you feel about something like
```
<html>
Welcome to ChartMuseum. To add this repository from the Helm CLI, run ...
</html>
```
?

liamawhite:
Sounds good to me!

PR #48, opened 2018-02-19
Redirect root route to /index.yaml, fix issue comparing Azure timestamps

jdolitsky:
Fix for issues #46 and #47

### DIFF
--- a/pkg/chartmuseum/logging.go
+++ b/pkg/chartmuseum/logging.go
@@ -109,7 +109,7 @@ func loggingMiddleware(logger *Logger) gin.HandlerFunc {
 		}
 
 		switch {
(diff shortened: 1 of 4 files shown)


liamawhite:
FYI: Helm CLI doesn't currently follow the redirect just 404s. 

See https://github.com/kubernetes/helm/issues/3531

davidovich:
I am having second thoughts about this. If the use case is user driven, could a simple html be served to help the user about the usage of the service ? This way chartmuseum keeps ownership of the format as opposed to deciding that the root will forever serve the index.yaml...

jdolitsky:
closing this for now until we agree on #46 . Going to open a new one that just fixes #47

PR #53, opened 2018-02-28
Multi-Tenancy POC

jdolitsky:
POC for enabling multi-tenancy. Do not merge. This is just meant to begin an open dialogue.

This introduces the popular [GORM](https://github.com/jinzhu/gorm) package to connect to temporary sqlite3 database "./gorm.db". GORM has support for mssql, mysql, and postgres as well.

GORM is an object relational mapper (ORM) library with syntax similar to Perl's DBIx::Class, which adds user-friendly methods and minimizes amount of raw SQL in the code.

Two database models have been created: Org and Repo. An Org has one-to-many Repos. A Repo is a ChartMuseum repo as we know it today.

Routes for working in/manipulating Orgs and Repos have been added under the temporary "/mt" prefix:
- `GET /mt/:org/:repo/index.yaml` - get index.yaml
- `GET /mt/:org/:repo/charts/:filename` - get chart packages etc.
- `GET /mt` - get all orgs
- `POST /mt` - create org
- `GET /mt/:org` - get org
- `DELETE /mt/:org` - delete org
- `POST /mt/:org` - create repo
- `GET /mt/:org/:repo` - get repo
- `DELETE /mt/:org/:repo` - delete repo

The routes above are all wrapped in `databaseMiddleware`, which add the the Org and Repo objects to the Gin context. If either are not found, we return a 404 and abort the request.

Things NOT covered in this POC:
- Custom database configuration
- Custom storage configuration per repo (should we even do this?)
- Caching of index.yaml per each repo. Currently hardcoded to look at `./charts/myorg/myrepo` dir on local filesystem, and generated upon every request
- ChartMuseum CRUD API per repo (e.g. `curl http://localhost:8080/mt/:org/:repo/api/charts`)

Open questions:
- How do we appropriately cache the index.yaml/chart metadata per repo? Need Redis?
- To prevent large number of `ListObjects` operations, should references to chart packages be stored in relational db as well? Do we rely on the db records instead of what's actually in storage?
- How can we enable RBAC on this? Can we start with just adding a `--disable-org-api` flag? How do we suggest people can plug this into their own internal auth system?
- Helm repos only support basic auth at this time. Should the auth mechanism for manipulating orgs, repos, charts, etc. differ from that of the helm repos themselves?
- Should ChartMuseum users still be able to use the original, database-less web server? Should multi-tenancy be enabled by default? Many users are satisfied with the current set of functionality

cc: @SlickNik @liamawhite @davidovich @mattfarina @itaysk @verchol @olegs-codefresh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@
 .test/
 .venv/
 bin/
+gorm.db
+gorm.db-journal
 goviz.png
 index.yaml
 mirror/
--- a/Makefile
+++ b/Makefile
@@ -28,13 +28,13 @@ build_windows:
 	    -o bin/windows/amd64/chartmuseum cmd/chartmuseum/main.go  # windows
 
 build_linux: export GOARCH=amd64
-build_linux: export CGO_ENABLED=0
+build_linux: export CGO_ENABLED=1
 build_linux:
 	@GOOS=linux go build -v --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
 	    -o bin/linux/amd64/chartmuseum cmd/chartmuseum/main.go  # linux
 
 build_mac: export GOARCH=amd64
-build_mac: export CGO_ENABLED=0
+build_mac: export CGO_ENABLED=1
 build_mac:
 	@GOOS=darwin go build -v --ldflags="-w -X main.Version=$(VERSION) -X main.Revision=$(REVISION)" \
 	    -o bin/darwin/amd64/chartmuseum cmd/chartmuseum/main.go # mac osx
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: dd0a48300d5bed94d76cfa166b2550aaf3da17f1cd8646c4bb1028aeabe2cf8d
-updated: 2018-02-19T17:22:05.874526-06:00
+hash: f4826202c9871b42228762658f7e0b61c79ca535022ddc1772b69d826b657b63
+updated: 2018-02-27T09:51:05.637384-06:00
 imports:
 - name: cloud.google.com/go
   version: 3137f1def9552929c7beb11f2ac7d2dd998e4040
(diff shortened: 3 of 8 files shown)


jdolitsky:
To test, start with a local directory layout such as
```
charts
├── myorg
│   ├── myrepo
│   │   └── nginx-ingress-0.9.3.tgz
```

Then run with something like
```
# remove sqlite3 database files (if desired)
rm -f gorm.db*

# run locally-built version of ChartMuseum
bin/darwin/amd64/chartmuseum --port=8080 --debug \
    --storage=local \
    --storage-local-rootdir=./charts
```

Create org: myorg
`curl -vL -d '{"name": "myorg"}' http://localhost:8080/mt | jq`

Create repo: myrepo
`curl -vL -d '{"name": "myrepo"}' http://localhost:8080/mt/myorg | jq`

Get all orgs:
`curl -vL http://localhost:8080/mt | jq`

Get org: myorg
`curl -vL http://localhost:8080/mt/myorg | jq`

Get repo: myorg/myrepo
`curl -vL http://localhost:8080/mt/myorg/myrepo | jq`

Get index.yaml
`curl -vL http://localhost:8080/mt/myorg/myrepo/index.yaml`

Get chart tgz
`curl -vLO http://localhost:8080/mt/myorg/myrepo/charts/nginx-ingress-0.9.3.tgz`

Delete repo: myorg/myrepo
`curl -vL -X DELETE http://localhost:8080/mt/myorg/myrepo | jq`

Delete org: myorg
`curl -vL -X DELETE http://localhost:8080/mt/myorg | jq`

liamawhite:
The API and directory structure looks good. A couple of points on the implementation and some of our other requirements:

- The directory structure already maintains which belongs to which org so there doesn't need to be an extra database for this state. 

- The actual authorization/ownership shouldn't need to be ChartMuseum's responsibility, it should be delegated to a separate application through an authorisation interface. This should be pluggable so the service owner could provide one that always says yes or write an OAuth plugin and plug that in.

- Some kind of abuse prevention/quota enforcer to stop a potential attacker from pushing large amounts of storage.

- An event stream in order to audit usage.

davidovich:
@liamawhite 

I agree with your points, especially the authorization delegation bullet. But there is a thing I am not sure about:

> The directory structure already maintains which repo belongs to which org so there doesn't need to be an extra database for this state.

Chartmuseum uses cloud storage in addition to the filesystem storage. I am not sure coupling to the structure offered by the filesystem backend is a good idea. On cloud storage, directories are typically not enforced (only cosmetic on the name), which would introduce additional complexity for those backends.

liamawhite:
The trade-off here is the complexity of those backends vs requiring the use of a database. It's probably a fair point to discuss though. As an FYI the point I put forward is how docker/distribution (registry) does things.

jdolitsky:
closing this for tracking purposes, but lets continue discussing here

PR #57, opened 2018-03-02
Add flag to opt-in to multitenancy (fixes #56)

### LINKED ISSUE #56: New multi-tenancy features should be opt-in
A flag should be added to the CLI that explicitly enables experimental new features and routing discussed in #53.

Unless this flag is specified, ChartMuseum will function as usual.

jdolitsky:
Flag name is `--multi-repo`. Feel free to suggest another name. Single root route defined for now.

cc: @liamawhite


pkg/chartmuseum/routes.go
```diff
@@ -41,3 +41,10 @@ func (server *Server) setRoutes(username string, password string, enableAPI bool
 		writeAccessGroup.DELETE("/api/charts/:name/:version", server.deleteChartVersionRequestHandler)
 	}
 }
+
+func (server *Server) setMultiTenancyRoutes() {
+	server.Router.GET("/", func(c *gin.Context) {
+		msg := "This ChartMuseum install is running in multi-repo mode."
```
liamawhite: Are we going with `multi-repo` or `multi-tenant`? Should probably be consistent everywhere either way.

liamawhite: Should also put a message saying that its pre-alpha and not currently expected to work.

liamawhite: Also is it worth having a separate package for the multi-tenancy code? As long as you have a standard interface it will keep it more modular and more readable.

jdolitsky:
Thanks for merge.

> Also is it worth having a separate package for the multi-tenancy code? As long as you have a standard interface it will keep it more modular and more readable

Yes, last commit prepares for that. Opened #63 to continue discussion.

PR #61, opened 2018-03-07
enable custom base path

rpahli:
enable custom base path to run under non root path


jdolitsky:
Cool thank you! Also, thanks for adding tests!

jdolitsky:
@rpahli this is now in v0.4.1

PR #67, opened 2018-03-09
Give root route same access at index.yaml (fixes #66)

### LINKED ISSUE #66: Root url not using authentication when API is
It seems like the root `/` isn't using authentication when the `/api` requires it. Is this expected?

I see there is work going on in #59 and #64 and I'm guessing this is vestigial from #29 when the root was the healthcheck.

Is there a reason we leave the `/` unauthenticated?

jdolitsky:
@liamawhite will this break your implementation?

cc @christopherhein


jdolitsky:
great, thanks!

PR #84, opened 2018-03-22
Move from glide to dep

### LINKED ISSUE #81: Use dep as package manager
The go community is moving to [dep ](https://github.com/golang/dep) as the official packaging solution.

glide is in [maintenance mode only](https://github.com/Masterminds/glide#golang-dep).

SlickNik:
Migrate from glide to dep since glide is now maintenance mode only and
dep seems to be what the community is coalescing on.

Fixes #81


jdolitsky:
Sweet thanks @SlickNik !

davidovich:
Tested locally, good job @SlickNik

PR #85, opened 2018-03-23
Multitenant repository index caching

### LINKED ISSUE #73: Multitenant server should use cache
Multitenant server is currently generating index.yaml upon every request, should be switched to use the cache

jdolitsky:
Added caching on a per-tenant basis, matching what is currently done for single tenant. The cache will only be pre-populated upon startup if depth=0 (for now). Fixes #73.

Special thanks to @davidovich for supporting the `--index-limit` flag on this.

Other small things:
- added verbose logging for new `dep ensure` command
- removed unneeeded code related to in-memory cache
- support for `--gen-index` and `--context-path` in multitenant server
- add GET /health route to multitenant server
- new Route type, which enables router to be only place where route transforms occur
- most robust middleware for parsing repo name out of path etc.


pkg/chartmuseum/server/multitenant/cache.go
```diff
@@ -0,0 +1,313 @@
...
+			"action", action,
+			"package", object.Path,
+		)
+		return nil
+	}
+	return err
+}
+
+func (server *MultiTenantServer) initCachedIndexFile(log cm_logger.LoggingFn, repo string) {
+	if _, ok := server.IndexCache[repo]; !ok {
```
davidovich: I think there is a race here. IndexCache could be evaluated empty for a given repo while another thread is initializing the repo cache.  Why not move the Lock() on function entry ? (then you could defer right away also)

Issue #93, opened 2018-04-03
Feature request: Allow different stores for different tenants


davidovich:
The current multitenancy implementation only allows one store per charmuseum instance.

We should allow multiple stores for multiple tenants. 

This brings another use for a central configuration file, as adding this through the command line will become unwieldy.

davidovich:
see #94

Issue #86, opened 2018-03-23
Should be able to specify different basic auth username/pass per multitenant repo


jdolitsky:
@goruha @itaysk we have chatted about this.
This will likely need to be fixed by introducing a config file and/or config database

davidovich:
See #94

PR #99, opened 2018-04-07
Ability to specify config via yaml file

### LINKED ISSUE #94: Feature request: Add a mechanism to specify configuration
The current implementation uses command line switches to control the configuration of the chartmuseum. This will become very hard to maintain and or implement if we want to support per tenant configurations.

We could add a configuration file to allow varying options for the chartmuseum startup.

jdolitsky:
Fixes #94. Is a prerequisite for #93 and #86.

Added the ability to specify a yaml config file using the `-c/--config` flag.

The order of configuration variable overrides is the following:

- config file
- environment variable
- command line flag

Uses the [viper](https://github.com/spf13/viper) package to parse the configuration. Viper allows for remote configuration in both etcd and Consul, which will become important to support users who want to change repo-specific config on-the-fly (public vs private repo etc).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -103,6 +103,12 @@
   version = "v3.1.0"
 
 [[projects]]
+  name = "github.com/fsnotify/fsnotify"
+  packages = ["."]
+  revision = "c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9"
+  version = "v1.4.7"
+
+[[projects]]
   name = "github.com/ghodss/yaml"
   packages = ["."]
   revision = "73d445a93680fa1a78ae23a5839bad48f32ba1ee"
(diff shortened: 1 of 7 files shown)


jdolitsky:
cc @itaysk @goruha

goruha:
@jdolitsky I think that using https://github.com/spf13/cobra can simplify code alot.
It have perfect integration with viper

jdolitsky:
@goruha I was looking at that as well. I think we probably should switch
from cli to cobra if we’re using viper.

So you think then that viper is the appropriate library to use here for
per-repo configuration? Is etcd/consul a good place to keep this info?

jdolitsky:
@goruha I've opened #100 for the switch. Feel free to take a stab :)

PR #110, opened 2018-04-23
Openstack support

gmauleon:


thelinuxfoundation:

* One or more commits are not associated with a GitHub user. The email used in the Git commit must match a verified email address in a GitHub user profile. For more information, please see https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

_Regards,
CLA GitHub bot_

gmauleon:
Fixed author e-mail

gmauleon:
Fixed Linux Foundation account

pkg/storage/openstack.go
```diff
@@ -0,0 +1,182 @@
...
+	// Create a custom HTTP client to handle reauth retry and custom CACERT if needed
+	roundTripper := ReauthRoundTripper{}
+	if caCert != "" {
+		caCert, err := ioutil.ReadFile(caCert)
+		if err != nil {
+			panic(fmt.Sprintf("Openstack (ca certificates): %s", err))
+		}
+
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
```
davidovich: Should we be droping the return value here ?

pkg/storage/openstack.go
```diff
@@ -0,0 +1,182 @@
...
+	var objects []Object
+
+	prefix = pathutil.Join(b.Prefix, prefix)
+	opts := &osObjects.ListOpts{
+		Full:   true,
+		Prefix: prefix,
+	}
+
+	pager := osObjects.List(b.Client, b.Container, opts)
+	page, err := pager.AllPages()
```
davidovich: Can we introduce a kind of iteration here ? What happens if there are thousands of charts ?

pkg/storage/openstack.go
gmauleon: Yeah, even if it does not return an error, I can throw a pem parsing error or the like myself indeed.

pkg/storage/openstack.go
gmauleon: Can sure do something like that, it's basically what the client is doing in AllPages() but I imagine it would be more memory efficient to do it on my side. Will look into that.

jdolitsky:
@gmauleon sorry for the delay on testing this! thanks so much for submitting!! will try to get to asap

gmauleon:
Roger that. Thanks :)

jdolitsky:
@gmauleon tested and works great! merged! will cut a release soon

Thanks again!

PR #125, opened 2018-05-10
Use default 'chart' and 'prov' form fields

### LINKED ISSUE #124: helm push uses "chart" as form field which could differ from chartmuseum configuration
We use the `CHART_POST_FORM_FIELD_NAME=chartfile` configuration option (to support legacy script clients using curl to upload charts). 

Right now, the helm push plugin hardcodes the form field to "chart" which does not allow it's usage with our CHART_POST_FORM_FIELD_NAME configured instance. Using `chart` is practical, as we do not want to expose this internal configuration to the helm plugin (or users), but having the knob to change the form field is also wanted. We can have both.

I propose that chartmuseum support the `chart` form field as a baked-in default (as a cli supported upload interface definition), and support additionally the CHART_POST_FORM_FIELD_NAME. In other words these two scenarios would be allowed:

* helm push will be able to use the default `chart` as supported by default.
* clients will be allowed to upload with `chartfile` as form field as configured by `CHART_POST_FORM_FIELD_NAME`.

davidovich:
Even if chartmuseum is configured with a form field through
CHART_POST_FORM_FIELD_NAME or PROV_POST_FORM_FIELD_NAME, also allow
posting with default 'chart' and 'prov' form field values.

This simplifies other clients like helm push which does not
know about the specific server setting.

closes #124

### DIFF
--- a/pkg/chartmuseum/server/multitenant/handlers.go
+++ b/pkg/chartmuseum/server/multitenant/handlers.go
@@ -12,10 +12,10 @@ import (
 )
 
 var (
-	objectSavedResponse        = gin.H{"saved": true}
-	objectDeletedResponse      = gin.H{"deleted": true}
-	healthCheckResponse        = gin.H{"healthy": true}
-	welcomePageHTML            = []byte(`<!DOCTYPE html>
+	objectSavedResponse   = gin.H{"saved": true}
+	objectDeletedResponse = gin.H{"deleted": true}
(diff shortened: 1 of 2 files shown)


jdolitsky:
David, I see the “WIP” title, is this ready to merge?

davidovich:
I want to give it a test run first :-), I'll remove the WIP when satisfied.

PR #127, opened 2018-05-15
Use statefiles to improve performance, limit GetObject requests

### LINKED ISSUE #91: Performance of indexer on startup
## Exists:
1. Chartmuseum backed by s3 bucket
2. There is ~500 charts in s3 bucket

## Operation:
Update [chartmuseum release](https://github.com/kubernetes/charts/tree/master/incubator/chartmuseum)

## Exists:
Startup takes too long

## Expect: 
Faster startup

## Possible solutions:
1. Use batch download to get `ObjectsList()` instead of `GetObject()`
2. Use lazy indexing (cache regeneration) on startup.
3. Use background indexing (cache regeneration) on startup.

jdolitsky:
Fixes #91 

A new file called `index-cache.yaml` is introduced, which, if found in storage, will be used to pre-populate cache on a per-tenant basis.

The source of truth is *still* the contents of storage (i.e. `ListObjects()`). The only real difference here is that upon request of index.yaml we compare the contents of `index-cache.yaml` to that of `ListObjects()`, and we only perform `GetObject()` calls based on the diff.

For example, if there are 100 packages in storage, but only 98 of them are listed in `index-cache.yaml`, the result is 2 `GetObject()` calls. Prior to this change, it was 100 `GetObject()` calls *no matter what*.

On the flip side, if there are 0 packages in storage, but there are 100 listed in `index-cache.yaml`, the contents of `index-cache.yaml` are essentially ignored, because there were no corresponding objects found in storage.

Also included in this PR:

- No longer saving 2 structs per tenant in cache, able to resolve diff between index and list of objects
- Improve acceptance test reliability, add function to wait for server to startup instead of sleep
- Upgrade Helm to 2.9.1
- Update `dep` dependencies


README.md
```diff
@@ -257,6 +257,7 @@ The contents of index.yaml will be printed to stdout and the program will exit.
 #### Other CLI options
 - `--log-json` - output structured logs as json
 - `--disable-api` - disable all routes prefixed with /api
+- `--disable-statefiles` - disable use of index-cache.yaml to improve caching
```
davidovich: not sure about the usage help: it *improves* caching to remove the statefile ?

jdolitsky: @davidovich thanks! addressed

davidovich:
Let me try this this morning. Otherwise outstanding work!

davidovich:
As I see it, the breakthrough of this PR is that it removes the need to keep a separate storage object list by using the modification date of a chart entry to build a fake storageObject ([LastModified: chartVersion.Created](https://github.com/kubernetes-helm/chartmuseum/pull/127/files#diff-e4a78696e7b46c8d2418e143d91eb293R77)), by which we can then calculate a diff. Excellent.

Tested in a staging environment, and works well.

Note however that as I said before, this does not relieve GetObject() pressure for the very first request. I think we can live with that until/if it becomes a problem.

jdolitsky:
@davidovich thanks for testing! 

and yes, youre correct about the pressure of GetObject  for very first request (per tenant)

Maybe later we add something that loads these on a timer... im not sure best approach. This is probably not the best approach, especially considering a large multitenant install with limited cache memory

Issue #128, opened 2018-05-16
Docs update request - Please add information regarding metrics


timstoop:
Hi,

I noticed from the chartmuseum Chart in kubernetes/charts that it is possible to gather metrics with Prometheus from chartmuseum. However, it is not explained anywhere how those metrics are actually exposed. Are they just exposed via /metrics on the main port? I'd love to see this updated in the docs. Unless I missed the explanation about it somewhere?

davidovich:
I'm working on a fix for Prometheus metrics, So I will take the opportunity to add some documentation. Thanks for the interest and suggestion.

timstoop:
Much appreciated!

davidovich:
Note that there was a change in the metric names and that the docs are slightly ahead of the current release. Please see #130.

PR #130, opened 2018-05-17
add tenancy metric dimension

### LINKED ISSUE #123: Prometheus monitoring is not multitenant aware
Metrics that were added to the museum only took one tenant into account, because there were no multitenant option at the time of it's inception. Currently, when we add multiple tenants, the per tenant index metrics get calculated, but overwrite the previous index metric calculation from a different tenant.

We could add the tenant dimension ([repo]) to the current metrics:  `charts_served` and `total_chart_versions_served`.

davidovich:
metric names are changed with this commit to be better
aligned with the prometheus naming best practices. They are renamed from

`total_charts_served` to `charts_served_total`
`total_chart_versions_served` to `chart_versions_served_total`

Fixes #123 and #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,8 +343,8 @@ have repo labels named `org1/repoa` and `org2/repob`.
 
 | Metric                                   | Type           | Labels     | Description                              |
 | ---------------------------------------- | -------------- | ---------- | ---------------------------------------- |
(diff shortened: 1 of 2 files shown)


jdolitsky:
@davidovich this is great, bonus point for tests

only comment is that "gauge" misspelled in readme

PR #132, opened 2018-05-22
Fix double PutObject and presence validation

### LINKED ISSUE #131: Double Put and object validation occur since we support default `chart` field
In #125 we added support for default `chart` post field.

This has a bad side effect of putting and validating the object twice when no field is configured for CHART_REPO_FORM_FIELD_NAME, which we should avoid as it is the default configuration.

davidovich:
In a recent change, it was made possible to have two form fields to post a prov or chart. This change introduced a double post and a double verification when the ChartPostFieldName is the same as the default one.

We fix this by using the filename to index into a map of chartOrProvenaceFile. This allows detecting only one contribution to each chart and prov files.

Fixes #131

### DIFF
--- a/pkg/chartmuseum/server/multitenant/handlers.go
+++ b/pkg/chartmuseum/server/multitenant/handlers.go
@@ -49,7 +49,7 @@ type (
 )
 
 type (
-	packageOrProvenanceFile struct {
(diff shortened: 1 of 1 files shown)


jdolitsky:
lgtm!

PR #133, opened 2018-05-24
Redis support

### LINKED ISSUE #72: Option to use Redis in place of in-memory cache
In support of multitenancy, structs that used to generate index.yaml should be able to be stored in Redis to tackle scale

jdolitsky:
Fixes #72 

The multitenant server type now has two new fields, `InternalCacheStore` and `ExternalCacheStore`.

If `server.ExternalCacheStore` exists, we make use of a new cache interface to get/set entries. This interface currently only has one implementation (which is Redis).

If `server.ExternalCacheStore` is nil, we continue to do caching like we always have.

cc @davidovich


pkg/chartmuseum/server/multitenant/cache.go
```diff
@@ -67,59 +60,64 @@ func (server *MultiTenantServer) primeCache() error {
...
+func (server *MultiTenantServer) regenerateRepositoryIndex(log cm_logger.LoggingFn, entry *cacheEntry, diff cm_storage.ObjectSliceDiff) <-chan indexRegeneration {
 	ch := make(chan indexRegeneration, 1)
 
-	server.IndexCache[repo].regenerationLock.Lock()
-	server.IndexCache[repo].regeneratedIndexesChans = append(server.IndexCache[repo].regeneratedIndexesChans, ch)
+	repo := entry.RepoName
 
-	if len(server.IndexCache[repo].regeneratedIndexesChans) == 1 {
-		server.IndexCache[repo].regenerationLock.Unlock()
+	server.Tenants[repo].RegenerationLock.Lock()
```
davidovich: Might as well take a reference to the tenant instead of indexing each time

pkg/chartmuseum/server/multitenant/cache.go
```diff
@@ -147,11 +145,36 @@ func (server *MultiTenantServer) regenerateRepositoryIndexWorker(log cm_logger.L
...
 
-	server.IndexCache[repo].RepositoryIndex = index
+	entry.RepoIndex = index
 
 	log(cm_logger.DebugLevel, "index.yaml regenerated",
 		"repo", repo,
 	)
+
+	// save to cache (either internal or external)
+	if server.ExternalCacheStore == nil {
```
davidovich: what about abstracting this cache work to a function ?

jdolitsky:
@davidovich comments addressed!

PR #135, opened 2018-05-29
Added --history-limit option.

aleerizw:
Old chart versions are deleted when a chart is added

Implements https://github.com/kubernetes-helm/chartmuseum/issues/134

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ The contents of index.yaml will be printed to stdout and the program will exit. 
 - `--index-limit=<number>` - limit the number of parallel indexers
 - `--context-path=<path>` - base context path (new root for application routes)
 - `--depth=<number>` - levels of nested repos for multitenancy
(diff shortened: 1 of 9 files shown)


aleerizw:
@davidovich Please let me know if there are any changes required. Thanks!

jdolitsky:
hi @aleerizw we will try to get this merged this week, thanks for your patience

davidovich:
@aleerizw sorry for the silence, work was keeping me very busy last week. 

I am a bit uneasy on the point of adding a deletion process to an add operation (uploadChartPackage).

Could we envision starting a goroutine at server startup time that collects older charts and does the house keeping? We could trigger/control the house keeping in a channel from the `uploadChartPackage` function.

What do you think ?

jdolitsky:
We do need some sort of background task management in general.

aleerizw:
I agree, I was not sure about this myself as well but wanted to get something out quickly :) I will make the changes

Issue #59, opened 2018-03-06
More robust authentication


itaysk:
Currently the Helm repo standard is to use anonymous HTTP or HTTP Basic authentication. This is a proposal to support a more robust authentication scheme with Chart Museum. Hopefully it will be embraced by Helm. If it won't, it still shouldn't affect Chart Museum's compliance with the Helm repo standard since this is an additional functionality on top of the standard functionality.  

> This was originated from the [discussion around Multi-tenancy to Chart Museum](https://github.com/kubernetes-helm/chartmuseum/pull/53), but it's a good feature for single-tenant users also hence this issue to kick-off this discussion as well. 

OAuth/JWT clarification:
OAuth2 is the common protocol for API authorization. It defines a set of scenarios (flows) that each describe how the client, server (resource server) and authorization server can work together to allow a client application to obtain an access token to some protected resources. The result of this auth dance is a token which the client can use for accessing the resources. The most common type of token is JWT, which is basically a signed JSON document with an agreed upon schema. When a user gets hold of a valid JWT (by following OAuth2, or by any other means), they will present it to the server which will validate the token, after which point the request is authenticated, and may proceed to making authorization decisions which are out of scope for now.

Chart Museum is a server that hosts charts. In OAuth2 lingo it is a resource server which rely on some other party to issue tokens. If we look at Docker Registry for reference, it seem to take similar approach of promoting OAuth2, but the open registry implementation (distribution) does not implement the entire flow (at least not as part of core), it expects the token to be issued by another component.

Taking this approach has the following benefits:
- Doesn't require a database (as requested by early adopters) but make it optional
- Cleanly integrate into hosting environments
- Leverage existing OAuth/JWT frameworks

To implement this, Chart Museum needs to accept JWT token:
- bootstrap CM using certain configuration like: trusted authorization server URL, valid signing keys, etc
- add a middleware that will enforce auth
- extract the token from the Authorization header
- validate the token
- examine allowed actions for this token (scopes)
- allow/disallow the request (disallow should return 401 with information on the required authentication)

References:
- OAUTH2 vs JWT: https://community.apigee.com/questions/21139/jwt-vs-oauth.html
- Docker Hub OAuth2: https://docs.docker.com/registry/spec/auth/token/
- Simple JWT validation as Gin middleware: https://github.com/gin-gonic/contrib/tree/master/jwt
- JWT verifivation details: https://auth0.com/docs/api-auth/tutorials/verify-access-token
[...]

PR #163, opened 2018-08-15
add RSA token validation

zachpuck:
As part of issue #59 this will allow chartmuseum to be configured to accept a token from an Authorization server (ex: [docker_auth](https://github.com/cesanta/docker_auth)) and verify its validity. The validation will only work for tokens signed with RSA at this time. 

To start using Bearer auth you will pass the following flags:
--bearer-auth=true
--auth-realm="<auth server endpoint>" 
--auth-service="<service name, ex: chartmuseum>" 
--auth-issuer="<name of auth server>"
--auth-cert-path="<path to auth server public cert pem file>"

Once added all endpoints will enforce auth. 

The validation will extract the token from the Authorization header, validate the token, and
allow/disallow the request (disallow will return 401 with responseHeader["WWW-Authenticate"] and the details for connecting to the Authorization server)

Currently we are not addressing Scopes on this PR and plan to complete that as a separate task.

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -92,7 +92,6 @@ func cliHandler(c *cli.Context) {
 		AuthService:            conf.GetString("authservice"),
 		AuthIssuer:             conf.GetString("authissuer"),
 		AuthCertPath:           conf.GetString("authcertpath"),
-		AuthScopes:             conf.GetString("authscopes"),
 	}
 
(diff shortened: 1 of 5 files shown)


jdolitsky:
Thank you! This is an amazing start. Works as advertised. Merging

Will put together a docker-compose example using `cesanta/docker_auth:latest` image. Repo placeholder here: https://github.com/chartmuseum/auth-server-example

PR #167, opened 2018-09-09
Run chartmuseum binary as nonroot user in container

verwilst:
Chartmuseum is currently running as root in the container. This conflicts with Kubernetes clusters that have PodSecurityPolicy support enabled.

Warning Failed 5s (x3 over 18s) kubelet, k8s-prod-worker02 Error: container has runAsNonRoot and image will run as root

This PR adds a user to the container and runs chartmuseum as it.


jdolitsky:
@verwilst great thanks!!

PR #171, opened 2018-10-05
Fixed openstack re-authentication

gmauleon:
Re-authentication attempts are counted throughout the whole life of the client.
In the case of a daemon like chartmuseum, this does not make sense at all since after 3 attempts the re-authentication will fail forever until the daemon is restarted.

This pull request make the re-authentication counter reset on successful calls, thus preventing a auth loop only on consecutive re-authentication.


jdolitsky:
@gmauleon lgtm. thanks!

Issue #173, opened 2018-10-15
wrong example in documentation for using s3 as a backend


trobert2:
Due to the latest change in the chartmuseum image here: https://github.com/helm/chartmuseum/blame/master/Dockerfile#L5
the root user is not the user used to execute the binary. it’s `chartmuseum`.

The docs state in the example of usage with S3 that the creds should be mounted under root.
```
docker run --rm -it \
  -p 8080:8080 \
  -e PORT=8080 \
  -e DEBUG=1 \
  -e STORAGE="amazon" \
  -e STORAGE_AMAZON_BUCKET="my-s3-bucket" \
  -e STORAGE_AMAZON_PREFIX="" \
  -e STORAGE_AMAZON_REGION="us-east-1" \
  -v ~/.aws:/root/.aws:ro \
  chartmuseum/chartmuseum:latest
```

Clearly wrong. After upgradint the path to `/home/chartmuseum/.aws` our issue was fixed.
An update to the documentation is required.

Thank you,
Robert

PR #174, opened 2018-10-15
Fixes S3 credentials volume attachment example.

trobert2:
Fix for issue #173


jdolitsky:
@trobert2 nice catch, thanks!

PR #207, opened 2019-01-18
migration to go modules - proposal

olegsu:
Relevant [builds](https://g.codefresh.io/public/accounts/olegs-codefresh/pipelines/olegsu/chartmuseum/chartmuseum) in Codefresh, using subset of steps from `.codefresh/master.yaml`
To make sure the builds are keep passing as long as test & acceptance tests.


davidovich:
I think this is the future. 

Can you explain why you kept the vendor dir but used it as a transient folder? If we need to download anyway, why build with vendor mode?

olegsu:
@davidovich , the vendor mode comes to make the CI process much faster.
In Codefresh each step is running in its own container.
So `SetupEnvironment ` step when running `make bootstrap` is doing `go mod download` and `go mod vendor`:
1. First download all dependencies into go `GOCACHE` folder (which is out of the working-directory)`
2. Now we need to bring them all into working directory. This is require becasue of how Codefresh is managing the shared volume between steps (only files under some path is part of the volume and `GOCACHE` isnt)

Now, when runing allbuild steps, if `-mod=vendor` is not set, go will try to download download all dependencies or get them from `GOCACHE` (which is empty), so `-mod=vendor` is manadatory to speed up the process.
To make sure nothing is exist before, `/vendor` is deleted in `make clean`.
Here's an [example](https://g.codefresh.io/public/accounts/olegs-codefresh/builds/5c41ef841c2e5a60086468b4) when running with `-mod=vendor` - build takes 3m.
And [without](https://g.codefresh.io/public/accounts/olegs-codefresh/builds/5c41e50a5f32ebc470a04a11) - build takes 6.5m

PR #209, opened 2019-01-26
Add step to set credentials for S3 compatible services

nitisht:
Signed-off-by: Nitish Tiwari <nitish@minio.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,11 @@ chartmuseum --debug --port=8080 \
   --storage-amazon-region="us-east-1"
 ```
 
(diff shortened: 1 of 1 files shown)


jdolitsky:
@nitisht does Minio not support creds via `~/.aws/credentials`? 

Using the env vars also applies to using S3 itself, so unless it is specific to Minio, I'd rather place this type of info closer to the top of the s3 section

nitisht:
> does Minio not support creds via ~/.aws/credentials?

@jdolitsky No, using `~/.aws/credentials` won't work out of the box for Minio deployments.

PR #214, opened 2019-03-15
Fix for multitenant repos with name api*

### LINKED ISSUE #201: 404 not found if calling /:repo/index.yaml when :repo=api
* set depth=1
* use repo name `api`

404 not found error will be returned when calling index.yaml of repo **api** ` /:repo/index.yaml (/api/index.yaml)` because the request will be identified as an API request. However, there is no API handler to handle it.

jdolitsky:
Fixes #201

Also in this PR:
- update dependencies
- add vanity import path "helm.sh" (to match https://github.com/helm/helm/pull/5440)


jdolitsky:
@steven-zou can you take a look at the fix? (the change is in match.go)

Basically if a route ends in .yaml/.tgz/.prov it will not be treated as an API route.

PR #228, opened 2019-05-03
update dependencies of chartmuseum

tariq1890:
i) This PR updates some dependencies to their latest versions.
ii) This also phases out `satori/go.uuid` in favour of `gofrs/uuid`. Satori's repo has not been maintained and lacks quite a few critical fixes.


pkg/chartmuseum/router/middleware.go
```diff
@@ -76,7 +76,7 @@ func setupContext(c *gin.Context) {
 	c.Set("requestcount", reqCount)
 	reqID := c.Request.Header.Get("X-Request-Id")
 	if reqID == "" {
-		reqID = uuid.NewV4().String()
+		reqID = uuid.Must(uuid.NewV4()).String()
```
jdolitsky: Can we remove the `Must` here? should be unnecessary, only used for things like
```
var packageUUID = uuid.Must(uuid.FromString("123e4567-e89b-12d3-a456-426655440000"));
```
according to https://godoc.org/github.com/gofrs/uuid#Must
`NewV4` should be good

tariq1890: I think we might need it, otherwise we would need to figure out how to bubble up the error that is return. `Must` swallows that error by panicking should the err be non-nil.

PR #248, opened 2019-06-19
Allow disabling delete route for a server

### LINKED ISSUE #243: Allow disabling the DELETE route
Presently, if the API is enabled, this also gives a user the DELETE permission globally. This can be a problem in multi-tenant installations as other tenants get the right to delete charts arbitrarily from other tenants.

We should decouple the DELETE route from the `APIEnabled` setting, in the same spirit as we allow controlling overwrites.

davidovich:
It may be desirable to limit destructive operations like the DELETE
route.

This change introduces another setting for disabling the destructive
delete route. It is inactive by default (hence the negated bool var
name), so it is backward compatible.

closes #243

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,7 @@ The contents of index.yaml will be printed to stdout and the program will exit. 
 - `--log-json` - output structured logs as json
 - `--log-health` - log incoming /health requests
 - `--disable-api` - disable all routes prefixed with /api
+- `--disable-delete` - explicitely disable the delete chart route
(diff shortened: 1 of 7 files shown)


thelinuxfoundation:

* One or more commits are not associated with a GitHub user. The email used in the Git commit must match a verified email address in a GitHub user profile. For more information, please see https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

_Regards,
CLA GitHub bot_

PR #262, opened 2019-07-30
fix etcd endpoint error and change readme.md example

guoweikuang:
fix etcd `endpoints` to `endpoint` and fix etcd example error


jdolitsky:
@guoweikuang thank you!

PR #265, opened 2019-08-23
add using yaml file example to readme

guoweikuang:
I found no examples of yaml configuration files in the documentation, which may be a bit bad for users who are just starting to use chartmuseum.


jdolitsky:
@guoweikuang thanks for this. Can you please sign the DCO? (see instructions above)

PR #282, opened 2019-10-26
sync API description with new API routes

mlycore:
changes like `/api/charts` to `/api/<repo>/charts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,15 +25,17 @@ Powered by some great Go technology:
 ## API
 
 ### Helm Chart Repository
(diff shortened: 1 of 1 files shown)


jdolitsky:
@mlycore Hi thanks for opening. 

While I do think this is helpful, I think it should be listed under the multitenancy section vs. at the top. This might be misleading for 90% of users who just want a single-tenant setup, especially without the context on how to use `--depth` / `--dynamic-depth` flags

jdolitsky:
@mlycore going to close for now. Perhaps we can revisit if we decide to make --dynamic-depth the default setting

PR #294, opened 2019-12-09
update to helm v3

### LINKED ISSUE #293: Unable to upload v2 charts
Hello, I'm using chartmuseum v0.10.0, and I'm not able to upload v2 charts. 

`{"error":"apiVersion 'v2' is not valid. The value must be \"v1\""}`

Does the current version of chartmuseum not support v2 charts?

karuppiah7890:
Fixes #293 

cc @jdolitsky


karuppiah7890:
@jdolitsky there was a build failure - https://g.codefresh.io/public/accounts/chartmuseum/builds/5dee68b6220dea7fb5bc7649

I think gocenter.io cannot be used when using `go` to build - https://jfrog.com/blog/oss-go-packages-in-gocenter/ . They recommend using `goc`

jdolitsky:
@karuppiah7890 lets disable gocenter for now.. we can add back later if needed

Anything with reference to MOD_PROXY_URL in Makefile can be removed

karuppiah7890:
@jdolitsky It's done 👍 How do I run the codefresh test again? It seems to run only once after raising PR. Only maintainers can rerun it?

jdolitsky:
@karuppiah7890 yes, theres some work to be done w/ CI. right now they trigger when label added

karuppiah7890:
the master branch failed - https://g.codefresh.io/public/accounts/chartmuseum/builds/5dee7b5ce1d2dd41c91fb016

ankushchadha:
@karuppiah7890 We removed goc dependency a while back when we started doing the 404 processing. The native go client can be used to pull modules from gocenter.

karuppiah7890:
I agree. Helm will fix the issue soon so nothing to worry about

ankushchadha:
Great, any reason why gocenter was removed as a GOPROXY? I would like to understand what didn't work, thanks.

karuppiah7890:
Could you check the above builds? Also there's an issue to add back GOPROXY. I'll do some experiments and add it back soon. Also idk if gocenter or proxy.golang.org will be chosen. I think adding a proxy URL with "direct" too should work. I was using golang:1.12 in which commas are not allowed and didn't notice build using chartmuseum/testbed image which has golang:1.13 in which GOPROXY can have commas. So nothing much over there. We'll add it soon I guess. :)

karuppiah7890:
Reason for using golang:1.12 in local - go mod tidy -v wouldn't work with golang:1.13 due to an issue in a dependency. You can check it out yourself

karuppiah7890:
But looks like that's not an issue in the builds or anywhere else when I noticed golang:1.13 being used and no tidying happening, so all good

ankushchadha:
Comma-separated GOPROXY values are only supported in Go 1.13, hence it won't work in Go 1.12.

The problem with only pointing to direct is that the builds won't be deterministic and it will be slow. With proxies such as GoCenter, we guarantee availability and immutability requirements to make the build process more deterministic. And with GOPROXY protocol, it's quick compared to pulling sources against github directly.

PR #314, opened 2020-02-25
Add flag for storage timestamp tolerance for invalidating cached index

### LINKED ISSUE #152: index.yaml regenerating every request even without any package updated
The index.yaml will regenerate every request to `/index.yaml` even though there is no any package changed.

When we compare the `LastModified` between objects from cache and storage:
https://github.com/helm/chartmuseum/blob/master/pkg/storage/storage.go#L64

The objects from cache is out of order due to map ranging....
https://github.com/helm/chartmuseum/blob/master/pkg/chartmuseum/server/multitenant/index.go#L107
```golang
func (server *MultiTenantServer) getRepoObjectSlice(entry *cacheEntry) []cm_storage.Object {
	var objects []cm_storage.Object
	for _, entry := range entry.RepoIndex.Entries {
		for _, chartVersion := range entry {
			object := cm_repo.StorageObjectFromChartVersion(chartVersion)
			objects = append(objects, object)
		}
	}
	return objects
}
```
And the `LastModifed` timestamp of object from storage seems has higher precision (I only test it on s3 backend)
```
LastModifed from cache: 2018-06-11T03:31:52.000Z
LastModifed from storage: 2018-06-11T03:31:52.094Z
```

Those above would cause the regenerating of index.yaml every request.

rajiteh:
Closes #152 
Related https://github.com/chartmuseum/storage/pull/39
Signed-off-by: Raj Perera <rajiteh@gmail.com>

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -65,6 +65,7 @@ func cliHandler(c *cli.Context) {
 	options := chartmuseum.ServerOptions{
 		StorageBackend:         backend,
 		ExternalCacheStore:     store,
+		TimestampTolerance:     conf.GetDuration("storage.timestamptolerance"),
 		ChartURL:               conf.GetString("charturl"),
 		TlsCert:                conf.GetString("tls.cert"),
 		TlsKey:                 conf.GetString("tls.key"),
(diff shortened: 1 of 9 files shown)


jdolitsky:
@rajiteh - thank you very much

PR #330, opened 2020-05-13
Remove redundant type conversion

hezhizhen:


cmd/chartmuseum/main.go
```diff
@@ -22,11 +22,11 @@ import (
 	"os"
 	"strings"
 
-	"github.com/chartmuseum/storage"
 	"helm.sh/chartmuseum/pkg/cache"
 	"helm.sh/chartmuseum/pkg/chartmuseum"
 	"helm.sh/chartmuseum/pkg/config"
 
+	"github.com/chartmuseum/storage"
```
jdolitsky, line 29: Can we put this import back w/ the other chartmuseum ones?

hezhizhen: Do you mean mixing imports from `helm.sh/chartmuseum` and `github.com/chartmuseum`?

hezhizhen:
@jdolitsky How do you group imports? I didn't find anything about it.

jdolitsky:
@hezhizhen - actually you're right, should be how you had it.. Dont have a documented process, just trying to follow same process in all files. I will go ahead and merge though, its not a big deal.

PR #338, opened 2020-06-10
chore(status code): use constant status code instead of raw

whalecold:
use the status code from pkg `net/http`


whalecold:
cc @jdolitsky

PR #339, opened 2020-06-10
Improve performance when dealing with thousands of charts

### LINKED ISSUE #332: Performance downgrade when dealing with thousands of charts
## Background
In our use case, it's easy to generate thousands of charts within couple weeks because CI/CD pipeline is integrated with hundreds of GitHub repositories and each pipeline run generates one chart.

Per observation, **It takes more than 10s to response any API (e.g. `GET /api/charts/<name>`) if there are about 5000 charts.**. Performance downgrade is dramatic and unacceptable. Especially when working together with Harbor, it makes the call to harbor frequently fail due to a fixed timeout interval (10s in this case).

## Analysis
All the `GET` requests call function `getIndexFile()`, which searches all the chart files from storage and updates the cache for `index.yaml` regardless whether the cache exists or not. It performs slower and slower with the growth of chart numbers. Although consistency and real-time are guaranteed, cache seems don't work at all.

## Proposal
To solve the problem and improve the performance, the only way is to make the cache "work" as expected. Therefore, I recommend to apply the following changes,
- Always read from cache in function `getIndexFile()` without traversing the files on the disk. If cache is missing, rebuild it
- Delta update the cache in all the WRITE requests (e.g. `POST /api/charts` or `DELETE /api/charts/<name>/<version>`)
- Delta update the cache every 5min (in case the files on the disk are manually manipulated)

A PR will be provided to fix the issue in 2 weeks. Hope some one from the governance team can review and accept it.

Ailsa-Wu:
## Summary
As we have disscsused in the [issue](https://github.com/helm/chartmuseum/issues/332), we need to improve the performance when dealing with thousands of charts. Therefor, we made some changes as we proposed before.

## Changes
Always read from cache in function getIndexFile() without traversing the files on the disk. If cache is missing, rebuild it
Delta update the cache in all the WRITE requests (e.g.` POST /api/charts` or `DELETE /api/charts/`)
Delta update the cache every 5min (in case the files on the disk are manually manipulated)
fixes #332


thelinuxfoundation:

* User @Ailsa-Wu isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

Ailsa-Wu:
retrigger

Ailsa-Wu:
retrigger

pkg/chartmuseum/server/multitenant/api.go
```diff
@@ -95,7 +95,7 @@ func (server *MultiTenantServer) deleteChartVersion(log cm_logger.LoggingFn, rep
 	return nil
 }
 
-func (server *MultiTenantServer) uploadChartPackage(log cm_logger.LoggingFn, repo string, content []byte, force bool) *HTTPError {
+func (server *MultiTenantServer) uploadChartPackage(log cm_logger.LoggingFn, repo string, content []byte, force bool) (string, *HTTPError ){
```
scbizu, line 98: Here will cause some compile problems since you not deal with the `error` branches ?

I think it's better to re-compile this PR after you rebased master ?

Ailsa-Wu: Oh,my fault. It was caused by my careless when I resolved conflict. I have fixed it. Thanks for your review again, and sorry for the problems!

Ailsa-Wu:
retrigger

Ailsa-Wu:
retrigger

Ailsa-Wu:
HI @scbizu , I have made all the check passed, may I have your review one more time ?  Appreciate it!

Ailsa-Wu:
hi @jdolitsky , I made some changes to improve the performance of Chartmuseum. It has worked well for about 3 weeks in our prod system. May I have a review? Hoping it is useful and can be merged to master. Thanks very much.

pkg/repo/index.go
```diff
@@ -99,6 +99,17 @@ func (index *Index) AddEntry(chartVersion *helm_repo.ChartVersion) {
...
+	entries := index.Entries[chartVersion.Name]
+	l := len(entries)
+	for i := 1; i <= 5 && l-i >= 0; i++ {
+		cv := entries[l-i]
+		if cv.Version == chartVersion.Version {
+			index.setChartURL(chartVersion)
+			entries[l-i] = chartVersion
+			return
+		}
+	}
```
jdolitsky, line 112: Could you help me understand what this is doing? What is the significance of the number 5?

cmd/chartmuseum/main.go
```diff
@@ -107,6 +109,18 @@ func cliHandler(c *cli.Context) {
...
+		for _ = range t.C {
+			server.(*multitenant.MultiTenantServer).RebuildIndex()
+		}
+	}()
+
+	go func() {
+		multitenant.InitUpdateObjectChan()
+		server.(*multitenant.MultiTenantServer).Consumer()
+	}()
+
```
jdolitsky, line 123: It appears that this modifies the default behavior, and that `RebuildIndex` will be run every 5 minutes on a timer (vs. on demand). Is that correct? 

I think that this should only occur based on the presence of some new flag, such as `--cache-interval=5m`.

These goroutines should also be encapsulated in some method on the server, such as `InitCacheTimer`

pkg/chartmuseum/server/multitenant/handlers.go
```diff
@@ -19,12 +19,17 @@ package multitenant
...
+	cm_logger "helm.sh/chartmuseum/pkg/chartmuseum/logger"
+	"helm.sh/helm/v3/pkg/chart"
 	"io"
 	"net/http"
 	pathutil "path"
 	"strconv"
-
-	"github.com/gin-gonic/gin"
+	helm_repo "helm.sh/helm/v3/pkg/repo"
+	"time"
```
jdolitsky, line 32: please see the imports order used in other files

pkg/repo/index.go
Ailsa-Wu: If someone push the same chart to `Chartmuseum` and `AllowOverwrite` is true, cache will  exist duplicate version info, like this
```yaml
  - created: "2019-10-21T09:46:09.216Z"
    digest: 963a506b5523966b993fe68c118d438402121b0d53cb412b9fc50328a428bc6e
    name: account1
    urls:
    - charts/account1-0.0.1-595d1a1bdb2d00edf976c46cde9c7562fab490ba.tgz
    version: 0.0.1-595d1a1bdb2d00edf976c46cde9c7562fab490ba
  - created: "2019-10-21T09:46:09.216Z"
    digest: 963a506b5523966b993fe68c118d438402121b0d53cb412b9fc50328a428bc6e
    name: account1
    urls:
    - charts/account1-0.0.1-595d1a1bdb2d00edf976c46cde9c7562fab490ba.tgz
    version: 0.0.1-595d1a1bdb2d00edf976c46cde9c7562fab490ba
```
It's a lazy solution to look for 5 versions from back to front. If none of the 5 versions are same to the new pushing version, I think it is a new version and not has been pushed. If there is a same version, I would update it. Because every new version will just be `append` to the end, I thought this solution is safe. Actually just comparing with the last one version is enough. For insurance purposes, I look for `5` versions from back to front and compare them with the new pushing version.

 ```go
index.Entries[chartVersion.Name] = append(index.Entries[chartVersion.Name], chartVersion)
```
I really hope you have a better way.

File: cmd/chartmuseum/main.go (hunk 3 above)
Comment (Ailsa-Wu) [reply cmd/chartmuseum/main.go:123]: Maybe the name `RebuildIndex`  is not
 very well, it delta update the cache every 5min (in case the files on the disk are manually manipulated).

It is a good idea to set a new flag and encapsulate these goroutines.

File: pkg/chartmuseum/server/multitenant/index.go [hunk 5]
```diff
@@ -39,49 +39,45 @@ func (server *MultiTenantServer) getIndexFile(log cm_logger.LoggingFn, repo stri
...
+			)
+			return nil, &HTTPError{http.StatusInternalServerError, errStr}
+		}
 
-	if server.UseStatefiles {
-		// Dont wait, save index-cache.yaml to storage in the background.
-		// It is not crucial if this does not succeed, we will just log any errors
-		go server.saveStatefile(log, repo, ir.index.Raw)
+		objects := server.getRepoObjectSlice(entry)
+		diff := cm_storage.GetObjectSliceDiff(objects, fo.objects, server.TimestampTolerance)
```
steven-zou, line 56: @Ailsa-Wu, @jdolitsky 

Will this method refer to the [latest improved version](https://github.com/chartmuseum/storage/blob/7feb4823f486829bd521ecf6ca31b9863601975e/storage.go#L57)?

steven-zou:
@jdolitsky 

When can this PR be merged and released? As you know, Harbor is depending on Chartmuseum to provide chart management capabilities and some of the Harbor users have met the serious perf issues of managing charts when there are large scale charts at the backend storage. Hope this PR can fix such perf issues of Chartmuseum. Looking forward to its merging and releasing.

scbizu:
Would this PR largely increases the memory that chartmuseum uses since we will maintain a large runtime cache? Should we cache all of the "thousands of"  charts ?

jdolitsky:
will take a look at this ASAP

Ailsa-Wu:
HI @scbizu ,  May I have your review one more time?

Ailsa-Wu:
hi @jdolitsky ,  may I have your review one more time?

pkg/chartmuseum/server/multitenant/index.go
jdolitsky: Just updated to latest storage release so yes 👍

jdolitsky:
@Ailsa-Wu - thanks so much for your patience. Finnally had a chance to look into this, and the change seems to work as expected. Merging.

@steven-zou - planning to cut a v0.13.0 release this week (thanks for Hrabor team patience as well)

jdolitsky (approved):

jdolitsky:
Please see docs on this new feature: https://github.com/helm/chartmuseum#cache-interval

mikhail-putilov:
@jdolitsky  hi :-) 

Do you have any news about planned 0,13 release? 

I am really anticipating this PR to be part of the next release.
Currently I am waiting 8 minutes to build a umbrella chart, it might speed up the building a lot

Regardless of the answer, I wanted to thank you for your contribution to open source, I really like chartmuseum :-)

jdolitsky:
@mikhail-putilov - I appreciate the kind words. Will try to get 0.13 out later in this week

peppe77:
@jdolitsky when will v0.13.0 be released? we have been experiencing this issue (#339), therefore heavily interested to upgrade to v0.13.0as it will hopefully solve the problem. thanks

djotanov:
I am interested as well when will the next version be relased?

Issue #308, opened 2020-01-21
Basic auth configuration does not seem to work via config file


jgoeres:
I would prefer using a config file instead of tons of command line args to configure Chartmuseum.
For our internal testing purposes, using basic auth with anonymous GET access should be fine, so I added these lines to my config file, 

    basic.auth.user: admin
    basic.auth.pass: s3cr37
    auth.anonymous.get: true

based on the documentation saying:

> When using a yaml file instead of the command line， Convert the - of the parameter in the command to .. Note that --storage is converted to storage.backend

Alas, it doesn't work, no authentication is active.

I also tried removing the "auth.anonymous.get" part, cause I was unsure what value to set here, since this is a command-line flag without an argument.

When I run chartmuseum and specify the basic auth options as command line args, it works as expected:

    ./chartmuseum --config chartmuseum.yaml --basic-auth-user admin --basic-auth-pass admin --auth-anonymous-get

Other settings from my config file (port, storage.backend etc.) are working fine.

I also tried leaving out 
But

jdolitsky:
@jgoeres it seems that the strings have an extra "." character. I realize this is misleading since the flags have "-" characters.

Please see all the keys here: https://github.com/helm/chartmuseum/blob/969515a51413e1f1840fb99509401aa3c63deccd/pkg/config/vars.go#L169

You should be able to do
```
basicauth.user: admin
basicauth.pass: s3cr37
```

Or alternatively:
```
basicauth: 
  user: admin
  pass: s3cr37
```

Please let me know if that works out

jgoeres:
Hi,
thanks, the basicauth.user and basicauth.pass keys work inthe config file, and in vars.go I found the equivalent of the switch "--auth-anonymous-get" to be 

    authanonymousget: true

Maybe the "mapping rule" mentioned in the docs that I quoted above should be removed entirely and replaced with a pointer to the vars.go file, on perspective to be replaced with a table of all command line flags and their corresponding config file key?

Regards

J

jayme-github:
[...]

PR #343, opened 2020-06-18
Register aliases for config keys that don't match the advertised schema

jayme-github:
This tries to clean up the situation an bit so that one can write
working configuration files without looking into the code.

I've added a disabled test to check for "invalid" config keys but that
would need renaming of the keys leading to incompatibility with current
contig files on user installations.

It would be nice to have a migration path, though.

Fixes: #308


jayme-github:
Just realized that the alias system of Viper does not work with config files as I had expected. This is more or less of no use then.

PR #345, opened 2020-06-23
Allow to bind chartmuseum to a specific interface

### LINKED ISSUE #255: Feature request: Bind to a specific IP address
**Use case**: I have chartmuseum run natively on my laptop (not in docker machine). However, from what I see, chartmuseum listens to all available network interfaces, thus any machine in the same wifi is able to access to my chartmuseum.

**Expect**: an argument in CLI to specify which network interface, aka. IP, to bind to.

jayme-github:
The new config option "listen.host" ("--listen-host", "LISTEN_HOST") may
be used to bind  chartmuseum to a specific interface rather than 0.0.0.0.

Default is 0.0.0.0 to stick with current behaviour.

Fixes #255

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -100,6 +100,7 @@ func cliHandler(c *cli.Context) {
 		WriteTimeout:           conf.GetInt("writetimeout"),
 		ReadTimeout:            conf.GetInt("readtimeout"),
 		EnforceSemver2:         conf.GetBool("enforce-semver2"),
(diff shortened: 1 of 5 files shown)


thelinuxfoundation:

* User @jayme-github isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

jayme-github:
> Can we modify to make this a field on the router, so that we do not need to modify the signature of router.Start or server.Listen?

Sure, let me know that you think.

PR #362, opened 2020-08-09
pkg/repo: inject chart metadata(name/version) into storage object

### LINKED ISSUE #220: Deleting a chart that has a version number in its name via the API causes no removal from index
It seems that deleting a chart that has a version number in its name via the API does not delete the entry from the index. The package is deleted from the file system. Seems similar to #22 

With a chart named `mychart`:
```
curl -X POST -k --data-binary "@mychart-0.1.1.tgz" localhost:8080/api/charts
{"saved":true}

curl localhost:8080/api/charts
{"mychart":[{"name":"mychart","version":"0.1.1","description":"my chart beta","apiVersion":"v1","appVersion":"1.0.0 BETA","urls":["charts/mychart-0.1.1.tgz"],"created":"2019-03-29T16:45:58Z","digest":"be1e1edef408420a3c004b2c88bd9c02ef3f34b6d346fe7a4a264297909da978"}]}

curl -X DELETE localhost:8080/api/charts/mychart/0.1.1
{"deleted":true}

curl localhost:8080/api/charts
{}
```

With a chart named `mychart-1.0`
```
curl -X POST -k --data-binary "@mychart-1.0-0.1.1.tgz" localhost:8080/api/charts
{"saved":true}

curl localhost:8080/api/charts
{"mychart-1.0":[{"name":"mychart-1.0","version":"0.1.1","description":"My Chart","apiVersion":"v1","appVersion":"1.0 BETA","urls":["charts/mychart-1.0-0.1.1.tgz"],"created":"2019-03-29T16:52:01Z","digest":"3a64e207bbc0c69e2027f7795b79c9feb5b64aa18c19e5730f95a09ae0cada57"}]}

curl -X DELETE localhost:8080/api/charts/mychart-1.0/0.1.1
{"deleted":true}

curl localhost:8080/api/charts
{"mychart-1.0":[{"name":"mychart-1.0","version":"0.1.1","description":"My Chart","apiVersion":"v1","appVersion":"1.0 BETA","urls":["charts/mychart-1.0-0.1.1.tgz"],"created":"2019-03-29T16:52:01Z","digest":"3a64e207bbc0c69e2027f7795b79c9feb5b64aa18c19e5730f95a09ae0cada57"}]}
```
Cheers,
Grant

scbizu:
e.g.: 0.0.1-SNAPSHOT-1 , it is a valid semver format but cannot be parsed properly (returns chart version 1 rather than 0.0.1-SNAPSHOT-1 )

Fix #220 

## Remark

I think I should add some more annotations here: 

Since deleting chart from index file should always delete the cache by chartname and chartversion , but the non-semver(or the chart not follow our parsing rules exactly) chart will be badly parsed the version with a incorrect number (as I comment in https://github.com/helm/chartmuseum/issues/220#issuecomment-669924807) . To resolve this issue , we should keep this chart version metadata within the all cache lifecycle , more closely , for the cached chart , we should add the metadata just after the cache inited (and store it in the storage.StorageObject.Metadata) , if so , the `diff` from cache and server storage will carry with the chart version , and then the `server.regenerateRepositoryIndex` will correctly regenerate the latest state of the server storage in cache.


kobynet:
@scbizu This will fix the problem with strict semantic versioning deletion, but the bigger problem is why to allow a chart creation that cannot be deleted? just like 1.0-SNAPSHOT+PR-102

scbizu:
> @scbizu This will fix the problem with strict semantic versioning deletion, but the bigger problem is why to allow a chart creation that cannot be deleted? just like 1.0-SNAPSHOT+PR-102

I think It's just a bug ? Or you ask why not use strict semantic versioning deletion before ? I don't know. Maybe chartmuseum itself allows the non-semantic charts ?

kobynet:
> > @scbizu This will fix the problem with strict semantic versioning deletion, but the bigger problem is why to allow a chart creation that cannot be deleted? just like 1.0-SNAPSHOT+PR-102
> 
> I think It's just a bug ? Or you ask why not use strict semantic versioning deletion before ? I don't know. Maybe chartmuseum itself allows the non-semantic charts ?

In your code at https://github.com/helm/chartmuseum/pull/362/files#diff-e4a78696e7b46c8d2418e143d91eb293R119
You check for strict semantic versioning, but the bug still exists for versions which are not strict semver just like 1.0-SNAPSHOT+PR-102 , we are not able to delete such versions even after your fix.

scbizu:
> > > @scbizu This will fix the problem with strict semantic versioning deletion, but the bigger problem is why to allow a chart creation that cannot be deleted? just like 1.0-SNAPSHOT+PR-102
> > 
> > 
> > I think It's just a bug ? Or you ask why not use strict semantic versioning deletion before ? I don't know. Maybe chartmuseum itself allows the non-semantic charts ?
> 
> In your code at https://github.com/helm/chartmuseum/pull/362/files#diff-e4a78696e7b46c8d2418e143d91eb293R119
> You check for strict semantic versioning, but the bug still exists for versions which are not strict semver just like 1.0-SNAPSHOT+PR-102 , we are not able to delete such versions even after your fix.

XD sorry for my misunderstanding . I will take a look at this , maybe we can have a solution with this case , but I think it's better for you to use a valid strict version rather than the no strict one.

kobynet:
> > > > @scbizu This will fix the problem with strict semantic versioning deletion, but the bigger problem is why to allow a chart creation that cannot be deleted? just like 1.0-SNAPSHOT+PR-102
> > > 
> > > 
> > > I think It's just a bug ? Or you ask why not use strict semantic versioning deletion before ? I don't know. Maybe chartmuseum itself allows the non-semantic charts ?
> > 
> > 
> > In your code at https://github.com/helm/chartmuseum/pull/362/files#diff-e4a78696e7b46c8d2418e143d91eb293R119
> > You check for strict semantic versioning, but the bug still exists for versions which are not strict semver just like 1.0-SNAPSHOT+PR-102 , we are not able to delete such versions even after your fix.
> 
> XD sorry for my misunderstanding . I will take a look at this , maybe we can have a solution with this case , but I think it's better for you to use a valid strict version rather than the no strict one.

I totally agree regarding having strict semantic versioning, was just wondering why fixing just the deletion of valid semver 2.0 fixed and not the other version :)

scbizu:
The previous code here is to find the proper version to delete but has the bad version detection for some cases , so I replace this code with the semver version detection for better readability and understanding.

(But I totally forget the non-strict semantic version :-| )

Maybe I should switch to another workaround.

jdolitsky:
@scbizu - can we prefix this PR with "WIP"?

kobynet:
@scbizu Any update on this issue?

scbizu:
@kobynet just finish my side after store the  chart version into Object metadata, and the PR is pending for merging https://github.com/chartmuseum/storage/pull/51 

If you are in hurry with this fix , could you self run the my patch , and check if it is ok : 

* pull my chartmuseum/storage branch (feature/object-meta) first 
* pull my helm/chartmuseum branch (fix/chartversion-int-after-dash) , the go.mod `replace` require you put `storage` and `chartmuseum` in the same directory.
* and then install the new museum with `make build-mac (or build-linux or build windows)`
* run the museum with `bin/darwin/amd64/chartmuseum --debug --port=8080 \
  --storage="local" \
  --storage-local-rootdir="./chartstorage"` to check if the issue is fixed

If you need a test docker image please at me , I think I can provide one.

jdolitsky:
That storage pr is now closed if you want to test here @scbizu

scbizu:
OK, I see it.  I will add some more tests  for this PR , thank u .

scbizu:
Finally all green , do you have time do some code review for me ? @jdolitsky

acceptance_tests/lib/ChartMuseum.py
```diff
@@ -89,6 +89,25 @@ def upload_test_charts(self):
                     self.http_status_code_should_be(201, response.status_code)
             os.chdir('../')
 
+    def upload_test_more_version_charts(self):
```
jdolitsky, line 92: I see these new functions here, but are they being used in acceptance tests? (.robot files)

jdolitsky (changes requested):

acceptance_tests/lib/ChartMuseum.py
scbizu: not yet . I forget it .

scbizu:
@jdolitsky hihi I re-edit the acceptance , now it seems right .

jdolitsky (approved):
This looks good. Will release new storage module, then we will update and merge

scbizu:
update the module dependency , but it seems like the dependabot will do it for me ? 🤔

jdolitsky (approved):

jdolitsky:
@scbizu thanks, we can close the bot PRs if they are out of date now

pixelicous:
@scbizu did you guys solve this issue? I still encounter it.. That problem and also the fact that even when deleting charts that do not exist that return deleted true

scbizu:
> @scbizu did you guys solve this issue? I still encounter it.. That problem and also the fact that even when deleting charts that do not exist that return deleted true

@pixelicous Did you bump to the latest version  ? Or can you provide us something more details ?

PR #363, opened 2020-08-09
[WIP]pkg/repo: changes the index entries map key from chart name to object path

scbizu:
Just a proposal:

If we can replace the index entries map key with object path , it will let us access the cache more convenient:

* No chart filename parsing
* Make the Object structure  more close to the ChartVersion structure 

But I wonder if these changes break the Helm spec about repo index entry itself ?

### DIFF
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -17,6 +17,8 @@ limitations under the License.
 package repo
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


jdolitsky:
@scbizu I accidentally closed this when changing target branch and cannot reopen... can you please open again? thanks!

Issue #391, opened 2020-11-22
Migrate release process to follow Helm's


jdolitsky:
Should adopt Makefile and release flow, and migrate from:

- Custom S3 bucket to get.helm.sh
- Docker Hub to GitHub registry

Releases should be signed, and should be able to install chartmuseum with install script similar to Helm's

jdolitsky:
Should adopt this document:
https://github.com/helm/helm-www/blob/master/content/en/docs/community/release_checklist.md

scbizu:
Hi @jdolitsky , really busy last month , and I have a short vocation these days . Can we continue with this ? 

Hmm , I read the [deploy.sh](https://github.com/helm/helm/blob/master/.circleci/deploy.sh) from helm , and do not the certain azure namespace or bucket for chartmuseum to store the binary . And also which the store `pattern` can we use ?

PR #395, opened 2020-11-29
ci: migrate release from docker hub into github container registry

scbizu:
Signed-off-by: scnace <scbizu@gmail.com>

Per #391 

- [ ] Custom S3 bucket to get.helm.sh
- [ ] Docker Hub to GitHub registry

### DIFF
--- a/.github/workflows/mig.yml
+++ b/.github/workflows/mig.yml
@@ -1,0 +1,82 @@
+name: mig-test
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: setup go environment
+        uses: actions/setup-go@v1
+        with:
+          go-version: '1.15.4'
+      - name: download dependencies
+        run: make bootstrap
+      - name: run unit tests
+        run: make test
+      - name: build binary
+        run: make build-linux
+      - name: run acceptance tests
+        run: sudo pip install virtualenv && make acceptance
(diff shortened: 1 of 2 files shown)


.github/workflows/mig.yml
```diff
@@ -0,0 +1,82 @@
+name: mig-test
+
+on:
+  pull_request:
+    branches:
+      - main
```
jdolitsky, line 6: Switch to build on push to main:
```
on:
  push:
    branches: main
    tags:
      - v*
```

GitHub actions will not use your the repo secrets on pull request, so we will just need to merge it to test

jdolitsky (changes requested):
I'm not sure we need personal access token.. but we can remove later

.github/workflows/mig.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: run unit tests
+        run: make test
+      - name: build binary
+        run: make build-linux
+      - name: run acceptance tests
+        run: sudo pip install virtualenv && make acceptance
+      - name: Prepare
+        id: prepare
+        run: |
+          DOCKER_IMAGE=chartmuseum/chartmuseum
```
jdolitsky, line 29: I think the issue is this probably needs to be `ghcr.io/helm/chartmuseum`

.github/workflows/mig.yml
scbizu: Sadly :(

.github/workflows/mig.yml
scbizu: okay i see

jdolitsky (approved):

PR #408, opened 2021-01-26
V0.13.0 prep

jdolitsky:
TODO:

- ~~re-enable docker build~~
- switch to `main` instead of `v0.13.0-prep` in installer script
- switch to `helm` instead of `jdolitsky` in installer script
- ~~gox build should pass in version+revision~~
- ~~some binaries build with v0.12.0 (vs. "canary")~~
- modify readme to remove aws references


jdolitsky:
closing in favor of #412

PR #415, opened 2021-02-09
fix release note sha256

houfangdong:
I found that the release package binary of s390x architecture is wrong，fix release sha256


jdolitsky (approved):

jdolitsky:
nice catch. will update the v0.13.0 release with correct sha

PR #423, opened 2021-02-23
pkg/chartmuseum/server: fix chart object path invalid

### LINKED ISSUE #421: Multi-tenant server is broken with local storage in v0.13.0
It looks like a change in `cache.go` to get the actual chart objects from storage when fetching the chart list forgot to join the repository path prefix to the object path like it should. The offending commit is ff22877341 and the lines in question are below:

https://github.com/helm/chartmuseum/blob/ff22877341b42b91efd3a4d1e5fcbab9a6eb7d91/pkg/chartmuseum/server/multitenant/cache.go#L204-L213

L207 should be doing `pathutil.Join(repo, object.Path)` like similar calls elsewhere in this and other files:

https://github.com/helm/chartmuseum/blob/ff22877341b42b91efd3a4d1e5fcbab9a6eb7d91/pkg/chartmuseum/server/multitenant/cache.go#L312-L317

I'd guess that this works fine with other storage providers and since they only seem to remove a prefix if it is present, they would have handled the missing prefix happily. But I'm commenting based on a short period of reading this code so it's definitely something worth looking into and smoke testing against more storage backends at least.

scbizu:
Fix #421 

- [x] Fix invalid path 
- [x]  Cover tests with multi-tenant server

### DIFF
--- a/pkg/chartmuseum/server/multitenant/cache.go
+++ b/pkg/chartmuseum/server/multitenant/cache.go
@@ -204,10 +204,12 @@ func (server *MultiTenantServer) fetchChartsInStorage(log cm_logger.LoggingFn, r
 	for _, object := range allObjects {
 		if object.HasExtension(cm_repo.ChartPackageFileExtension) {
 			// Since ListObject cannot fetch the content from file list
-			object, err = server.StorageBackend.GetObject(object.Path)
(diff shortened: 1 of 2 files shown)


scbizu:
Can we merge it now ? And release the v0.13.1 version or port it to 0.14.0 ? @jdolitsky

PR #454, opened 2021-05-03
pkg/chartmuseum/server: upload chart should emit updateChart event with overwrite option is set

### LINKED ISSUE #450: Duplicate versions for same chart
We are seeing many duplicate entries in `index.yaml` for the same chart version.  In some cases there are thousands of duplicates for the same version.  We expect only one entry per version/tag.

ChartMuseum version: 0.13.1
Running as part of Harbor with multiple instances of chartmuseum and settings:
```
ALLOW_OVERWRITE=true
CACHE: redis
DISABLE_STATEFILES: "false"
STORAGE: amazon
```

Is there a concurrency issue that we should be aware of?

Example:
```
  - apiVersion: v1
    created: "2021-04-10T14:59:57.826535265Z"
    digest: 1cd815901d2166307de13a7632ae04c053256863e486276c3411b26f7a3b2f0c
    name: ourrepo
    urls:
    - charts/ourrepo-0.0.0-84ace941.tgz
    version: 0.0.0-84ace941
  - apiVersion: v1
    created: "2021-04-19T08:59:10.291630412Z"
    digest: 1cd815901d2166307de13a7632ae04c053256863e486276c3411b26f7a3b2f0c
    name: ourrepo
    urls:
    - charts/ourrepo-0.0.0-84ace941.tgz
    version: 0.0.0-84ace941
```

scbizu:
Try to fix #450

### DIFF
--- a/pkg/chartmuseum/server/multitenant/api.go
+++ b/pkg/chartmuseum/server/multitenant/api.go
@@ -106,11 +106,18 @@ func (server *MultiTenantServer) uploadChartPackage(log cm_logger.LoggingFn, rep
 		return filename, &HTTPError{http.StatusBadRequest, fmt.Sprintf("%s is improperly formatted", filename)}
 	}
 
-	if !server.AllowOverwrite && (!server.AllowForceOverwrite || !force) {
-		_, err = server.StorageBackend.GetObject(pathutil.Join(repo, filename))
-		if err == nil {
+	// we should ensure that whether chart is existed even if the `overwrite` option is set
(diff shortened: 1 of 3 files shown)


jdolitsky (approved):
👍

ninjadq:
I think I got the reason why this will duplicate still exist. The harbor replication job replicates charts with the `prov` filed in request. So it go to another path in the code, and in that path, the issue didn't fix. I've opened a new PR to fix it

PR #464, opened 2021-06-07
[WIP] Add support for Metadata

ryanm101:
## Background
It's useful as part of a build/release process to be able to store metadata about a particular chart. For example you might have a chart that builds successfully but fails integration tests, this is still a valid build / chart but is unlikely to be releasable. 

Currently there is no way to tell the state of a chart in CM, this adds a simple file in the spirit of `.prov` called `.meta`. 
With the format:
```
{
  "name": "mychart",
  "version": "0.1.0",
  "somekey": "someval",
  "tests_passed": true,
  "override_deploy": true,
}
```

`curl --data-binary "@mychart-0.1.0.tgz.meta" http://localhost:8080/api/meta`

The only two reserved K/V pairs being `name` and `version` which must match the chart that the meta data is associated with. (Maybe `creation_ts` and `last_update_ts` needs to be reserved too???)

Opening this WIP now as others may have ideas for better implementation or comments on current process. 

Dummy Usecase 1:
1 . Build docker containers
2. Generate Helm Chart and upload to CM
3. deploy helm chart to some env
4. run acceptance / integration tests
5. update mychart-0.1.0.tgz.meta with `"tests_passed": true`
6. Some deployment process pulls chart.meta and checks K/V pairs for matches
7. Deployment process takes action depending on K/V pair E.g. Deploy or Not

## Limitations of this implementation
* Customers must write there own search / parser CM just hosts the metadata it's up to the customer to process it. Perhaps a later addition will be to let this be searched 
* Not easily searchable - I had thought sticking this in redis etc would be a better way and maybe long term  or for more advanced setup a build meta data server could be useful but this is a easy start that lets less advanced setups not worry about that complexity.

### DIFF
--- a/pkg/chartmuseum/server/multitenant/api.go
+++ b/pkg/chartmuseum/server/multitenant/api.go
@@ -190,6 +190,40 @@ func (server *MultiTenantServer) uploadProvenanceFile(log cm_logger.LoggingFn, r
 	return nil
 }
 
+func (server *MultiTenantServer) uploadMetaFile(log cm_logger.LoggingFn, repo string, content []byte, force bool) *HTTPError {
+	filename, err := cm_repo.MetaFilenameFromContent(content)
+	if err != nil {
+		return &HTTPError{http.StatusInternalServerError, err.Error()}
+	}
+
+	if pathutil.Base(filename) != filename {
+		// Name wants to break out of current directory
+		return &HTTPError{http.StatusBadRequest, fmt.Sprintf("%s is improperly formatted", filename)}
+	}
+
+	if !server.AllowOverwrite && (!server.AllowForceOverwrite || !force) {
+		_, err = server.StorageBackend.GetObject(pathutil.Join(repo, filename))
+		if err == nil {
+			return &HTTPError{http.StatusConflict, "file already exists"}
+		}
+	}
+	limitReached, err := server.checkStorageLimit(repo, filename, force)
+	if err != nil {
+		return &HTTPError{http.StatusInternalServerError, err.Error()}
+	}
+	if limitReached {
+		return &HTTPError{http.StatusInsufficientStorage, "repo has reached storage limit"}
+	}
+	log(cm_logger.DebugLevel, "Adding Meta file to storage",
+		"meta_file", filename,
+	)
+	err = server.StorageBackend.PutObject(pathutil.Join(repo, filename), content)
+	if err != nil {
+		return &HTTPError{http.StatusInternalServerError, err.Error()}
+	}
+	return nil
+}
+
 func (server *MultiTenantServer) checkStorageLimit(repo string, filename string, force bool) (bool, error) {
 	if server.MaxStorageObjects > 0 {
 		allObjects, err := server.StorageBackend.ListObjects(repo)
--- a/pkg/chartmuseum/server/multitenant/handlers.go
+++ b/pkg/chartmuseum/server/multitenant/handlers.go
@@ -293,6 +293,26 @@ func (server *MultiTenantServer) postProvenanceFileRequestHandler(c *gin.Context
 	c.JSON(201, objectSavedResponse)
 }
 
+func (server *MultiTenantServer) postMetaFileRequestHandler(c *gin.Context) {
+	repo := c.Param("repo")
+	content, getContentErr := c.GetRawData()
+	if getContentErr != nil {
+		if len(c.Errors) > 0 {
+			return // this is a "request too large"
+		}
+		c.JSON(500, gin.H{"error": fmt.Sprintf("%s", getContentErr)})
+		return
+	}
+	log := server.Logger.ContextLoggingFn(c)
+	_, force := c.GetQuery("force")
+	err := server.uploadMetaFile(log, repo, content, force)
+	if err != nil {
+		c.JSON(err.Status, gin.H{"error": err.Message})
+		return
+	}
+	c.JSON(201, objectSavedResponse)
+}
+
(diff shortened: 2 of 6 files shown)


scbizu:
Thanks for your idea and pr . But I think it is maybe not a good idea to store this kind of  "metadata" inside a separate file . 

We already stored the metadata inside our storage struct [`storage.Object`]() , but the kvs you listed are differs from users , It's better to let users to decide this . And CM should not care which chart is currently deployed , and which not . The version or tag is enough. 

However, it is still my personal opinion . We can talk more about your requirement if we really have the strong intention to do this 🙈

ryanm101:
The issue I'm trying to solve is that in a CI/CD pipeline we will build our Container, then we wrap that in a helm chart which then gets stored in CM, this object is immutable/versioned as it should be.
This creates an issue we now have an immutable object but we need some other storage medium to hold data about the Chart itself, we cant use chart comments "-xyz" as all builds should be deployable without being rebuilt and if we add a such a tag to the chart name it become nonsical e.g. in the case of an RC that we decide is actually what we want to release we don't want our 'production' chart to have a "-rc" in the name nor do we want to do a new build just to generate a clean release name.

If you look at the likes of nexus or artefactory when an object is uploaded there is the ability to store metadata about that object that can be added/removed/updated separately to the object. This allows deployment systems to query the metadata to allow them to make decisions. 
I could build a whole database for this purpose but this would mean the data is decoupled from the object and would require building some sort of integration system to update both CM and the DB storing the metadata. Use the method here of a seperate .meta file we neatly side step adding complexity to build systems or adding extra systems in to manage this metadata. 

Some examples of use cases below:
1)  Chart 1.0.0 is built and is a  release candidate instead of storing a chart named myapp1.0.0-rc then if we decided the chart is ti be released to prod we dont need to rebuild the chart to get a new name of myapp1.0.0. IMO this is important as if you rebuild you are effectively releasing a different chart than the one you just tested (and you would have to run through the whole build / test cycle again for nothing more than a name change. With the meta file we would just build myapp1.0.0 and provide this file which could contain a K/V such as 'isRC:True' and 'isReleasable:false' or whatever that could be used by the other systems to determine what then can do with that chart. 

2) A chart is built and uploaded but fails the qa tests the .meta file would allow us to add that info to the chart .meta file to prevent deployment systems being able to release it or it would allow  for us to query all charts for passes / fails.

scbizu:
Sorry there is a long time not responding .

Through your proposal , I know your requirements and also these are normal backgrounds for CD System , but I would say CM is just a Chart Management System , it stores chart that developers pushed and provides the interface to pull down . The `.meta` magic you described upon should be a part of the Chart . 

If you do not store these metadata in the Chart , the deployment will be different from CM and no-CM instance (Since Helm do not recognise  your `.meta` file). The storage metadata should always be synced with upstream Helm Chart metadata .

> if we decided the chart is ti be released to prod we dont need to rebuild the chart to get a new name of myapp1.0.0

What is the "rebuild" means ? Per my understanding , you just need to tag the tested version into the prod version , and then update the Chart's `.values.yaml` to point to the new docker image , and package the Helm Chart to store in CM . If your rebuild that , I think it is necessary and the common way for CD pipeline .

BTW , thanks for your such long proposal  : )

ryanm101:
> If you do not store these metadata in the Chart , the deployment will be different from CM and no-CM instance (Since Helm do not recognise your .meta file). The storage metadata should always be synced with upstream Helm Chart metadata .

The logic here makes no sense to me, as you are either using CM in which case you have the benefit of `.meta` or you are not in which case you don't and have to use a different solution. Additionally this is kind of the point in that it is meta data and not valid kubes / chart data. 

 > What is the "rebuild" means ? Per my understanding , you just need to tag the tested version into the prod version , and then update the Chart's .values.yaml to point to the new docker image , and package the Helm Chart to store in CM . If your rebuild that , I think it is necessary and the common way for CD pipeline .

This is probably more down to the way our pipelines are designed but we build an image that gets stored in Docker Registry, if that build is successful the next stage of the build generates a chart which is stored in CM. We also have a job that creates a chart of charts to manage interdependency between microservices. The only way (for us) to rename / retag a chart is to run our whole build process again and to change the tag (to remove -rc for example) this seems overkill. The other risk is this introduces is if something uses an unpinned downstream dependency (I have seen this with sensible where it used some lib that had dependences that didnt pin their dependences that then broke some of my scripts)  or even if the build image gets updated (new version of say gcc) then you could end up with a different binary unintentionally.

I just see this as a value add, and as I said before I could write a DB that stores all this meta data separately but that is a load of development work that doesn't couple the data with the subject and means we need some way of interacting with that data as part of the build process.

Anyhow if you feel there is no value in this, please feel free to just close out the PR.

Regards,

Ryan

scbizu:
Thanks for your explanation , Ryan .  I think this  proposal seems too early for the current CM , and maybe we can introduce the new `plugin` mechanism for CM which do not include by default and maintained not in the main CM repository , such as your `.metadata` . However , we should make this decision with the whole community , and discuss about how to work out this .

I will close this PR for now , and could you please let this PR to be an issue to track your proposal , and link back to this PR ?

Thanks ,

Nace

scbizu:
Thanks for your explanation , Ryan .  I think this  proposal seems too early for the current CM , and maybe we can introduce the new `plugin` mechanism for CM which do not include by default and maintained not in the main CM repository , such as your `.metadata` . However , we should make this decision with the whole community , and discuss about how to work out this .

I will close this PR for now , and could you please let this PR to be an issue to track your proposal , and link back to this PR ?

Thanks ,

Nace

Issue #316, opened 2020-03-20
Auto-purge, get rid of old chart versions


jdolitsky:
Add feature flags that enable auto-removal of old chart versions in storage based on various age / last used / version parameters

cep21:
I'm interested in this ticket.  It's something I could do out of band, but the missing piece is I don't think there's anything in chartmuseum that tracks last used for charts.  Does this data currently exist?  If we were to add it, where would it be stored?

scbizu:
@cep21 Happy with your interests , maybe [this](https://github.com/chartmuseum/storage/blob/main/storage.go#L32) is what you find ? If you can help us with this feature , it will make a big sense :)

cep21:
@scbizu I think "last read" makes more sense than "last modified", right?  You would want to remove charts people are no longer downloading, for example.

scbizu:
@cep21 You're right ,  but the storage itself do not support the last read timestamp yet . It will be huge PR if we add a new mechanism to the storage structure , we should handle every REQ from `helm pull`, and update the last read timestamp. 

I think this is why this issue still tag with `help wanted` XD

cep21:
One idea is to store this information inside redis, for example.  Another is to use the backend storage itself to store this information and "sync" some kind of ledger every 60 seconds (for example).

scbizu:
Maybe the AutoPurge should be provided as an interface function , and it can be differ from the real storage backends. The expiration duration should be configurable if users open the auto purge feature .

cep21:
> and "sync" some kind of ledger every 60 seconds 

This could be tricky if people are running multiple chartmuseum instances for redundancy, since we'll have to merge ledgers

vadasambar:
It'd be great if we could have more than one conditions to decide whether to delete the charts or not. E.g., Instead of "Delete all the charts older than 2 months", it would be better if we have "Delete all the charts older than 2 months matching a particular regex". This is because you might not want to delete release (e.g., `2.1.0`) charts but if you want to delete pre-release charts (e.g., `2.1.0-custom-fix` or `2.1.0-pr-3245`), you could use regex to match all the pre-release charts older than specified time period (Check #383 ). 

However, one thing that concerns me about regex is you'd want to test it out first to see which charts would be deleted with that particular regex to avoid deleting charts you didn't intend to delete.

scbizu:
The thread is too long to track information now . And let me draw the conclusion till now , the key points list below:

[...]

PR #466, opened 2021-06-14
Add per chart limit option

### LINKED ISSUE #376: Is there a way to limit number of old charts?
Is there currently any way to limit number of old charts cumulating over time?

There was some idea to do this back in 2018: https://github.com/helm/chartmuseum/pull/135 but looks like it was never finished.

scbizu:
impls #316 

Also , closes #376 

- [x] backend storage and index cache supports per chart limit

Maybe in next release:

-  lock chart to be unremovable
-  provide a dry run command to clean museum(per chart) , also works with chart lock mechanism

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -104,6 +104,7 @@ func cliHandler(c *cli.Context) {
 		EnforceSemver2:         conf.GetBool("enforce-semver2"),
 		CacheInterval:          conf.GetDuration("cacheinterval"),
 		Host:                   conf.GetString("listen.host"),
+		PerChartLimit:          conf.GetInt("per-chart-limit"),
 	}
 
 	server, err := newServer(options)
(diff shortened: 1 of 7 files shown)


scbizu:
I found a problem during implementing this feature , dose all of our storage impls support the same name file covering. With the #454 , we can now update our cache index in the chart updating case , but it still can be stored into backend storage if this backend storage do not support the same name file covering or add some other more suffixes into the package name . 

However, we maybe break the above case and the `per-chart-limit` option will also limit the same name artifact at N versions . I do not know if this should be mentioned in the next release note.

jdolitsky:
@scbizu - where do we stand with this and #457 ?

jdolitsky:
@scbizu can you rebase?

scbizu:
done @jdolitsky

jdolitsky (approved):

Issue #430, opened 2021-03-10
Unit and acceptance tests broken upgrading to Helm v3.5.(2?/3?)


jdolitsky:
The semver validation appears to be strictly enforced using `loader.LoadArchive`

This line causes 500: https://github.com/helm/chartmuseum/blob/main/pkg/repo/chart.go#L52

cc @scbizu

scbizu:
hmm ... It seems that the broken test caused by the path error , and the `testdata/badcharts/mybadsemver2chart/mybadsemver2chart-0.x.x.tgz` archive file is missing

scbizu:
It is weird that only the mybadsemver2chart not be packaged , maybe re-run makes sense ?

scbizu:
And I found why , the fork version https://github.com/bloodorangeio/chartmuseum/tree/0-13-1-prep/testdata/charts do not have mybadsemver2chart dir . Can you fix it @jdolitsky ?

jdolitsky:
@scbizu - the issue is the test setup script was not building the .tgz. I fixed this in this commit https://github.com/helm/chartmuseum/commit/65f7482f9a49b589652a2f7141459af0941a006f

Now the error is this:
```
    server_test.go:683: 
        	Error Trace:	server_test.go:683
        	Error:      	Not equal: 
        	            	expected: 400
        	            	actual  : 500
        	Test:       	TestMultiTenantServerTestSuite/TestSemver2Validation
        	Messages:   	400 POST /api/charts bad semver validation
```
        	
https://github.com/helm/chartmuseum/pull/429/checks?check_run_id=2091220693

The 500 is thrown because helm libraries will not parse the binary content into a chart due to strict semver check

scbizu:
I see , the Helm latest helm [commit](https://github.com/helm/helm/commit/657ce552cb6e582976c08cccc9605e42c242084e) forces the semantic version . For me . I think it is a good approach , it means that we can deprecated the  `--force-semver2` flag , and let it depend on the helm's validation.

But since v0.13.1 is a bug fix release , we should modify our go.mod to depend on the lower version of helm `pkg/chart/loader` to make our action pipeline happy and make this break changes in 0.14.0 maybe ?

jdolitsky:
@scbizu sure, that sounds good. just updated #429

PR #477, opened 2021-08-18
Support Builds for ARM M1 Macs

cfstras:
## Summary
This PR adds build targets to the makefile to support building for darwin/arm64, which is the architecture for Apples processors going forward from the M1.

It also adds support for the arch in the get-chartmuseum script, which will then attempt to download the version. As soon as the first version with this build is released, everybody can enjoy it :)

Also added `.PHONY` labels to the other build-* targets, to prevent make confusion when files called build-* exist.

## Testing

Successfully tested on an M1 MBP, macOS 11.5.2.

## Caveat

Unfortunately, there is a cosmetic error message at the start coming from containerd, because it tries to call `getCPUInfo()` on every ARM arch:
```
ERRO[0000] failure getting variant                       error="getCPUInfo for OS darwin: not implemented"
```
See https://github.com/containerd/containerd/blob/v1.3.4/platforms/cpuinfo.go#L77-L93.

The bug is already fixed since 1.4.0 of containerd/containerd: https://github.com/containerd/containerd/commit/7119a2a152edcf517a6cbc3271794c720bff2bd8, but indirect dependencies still pull in old versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,8 @@ BINNAME     ?= chartmuseum
 # Required for globs to work correctly
 SHELL      = /usr/bin/env bash
 
-TARGETS     := darwin/amd64 linux/amd64 linux/386 linux/arm linux/arm64 linux/mips64le linux/ppc64le linux/s390x windows/amd64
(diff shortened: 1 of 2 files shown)


scbizu:
Good job , could you pass the DCO verification ?

cfstras:
@scbizu Should be good to go :)

scbizu:
The containerd dependency is from the helm dependency though ...

However , we can not bump our helm dependency to the latest version due to [one potential break change](https://github.com/helm/chartmuseum/issues/430) .

And so , this PR may should be merged after we fixed the caveat ?

(After v0.14.0 ,  we need to deprecate the invalid semantic version by default and bump our Helm dependency to the latest and then we can also bump the containerd dependency to fix the caveat).

cfstras:
I don't you need hold this back -- as mentioned, the error message is purely cosmetic, and has no implications on chartmuseum itself (apart from being slightly annoying).

Being able to build this on M1 with a weird log entry is better than not being able to run it at all ;)

scbizu:
Sounds right , I can merge this in and it (the weird error message ) will disappear after we bump the containerd dependency , right ?

cfstras:
Correct, that should be the case. (noticed that the build failed though, it looks like helm does not live at https://storage.googleapis.com/kubernetes-helm/ anymore, but here: https://github.com/helm/helm/releases/tag/v2.17.0

scbizu:
Thank you , I will merge it and then fix the CI 😆

Issue #485, opened 2021-09-06
Address the incoming potential breaking changes since v0.14.0


scbizu:
## Background 

Since Helm do not support non-semantic  version after their `v3.5.1` release , If we need to follow with the Helm version , we must also deprecate the non-semantic version support . We have introduced the `enforce-semver2` since ChartMuseum v0.13.0 release , with this option set , we also force semver2 chart .

## Decision

With Helm's officially banned non-semver2 Chart , we decide to deprecate the `enforce-semver2` option , the option arg will keep , but it does nothing except the DEPRECATION WARNING . And , since chartmuseum v0.14.0 , the semver2 validation will depend on Helm's , and can not disable any more .

## The notice for different kind users

* For users which  need to store non-semver2 charts : Stay at `v0.13.1 ` , but please do the migration ASAP , we'd not make any release based on the non-semver2 charts.

* For users without setting `enforce-semver2`: Check your customer charts version , and make sure they are all valid semver2 charts. If not , go to usercase#1 , otherwise, go to usercase#3 .

* For users already set `enforce-semver2`: Just go ahead to `v0.14.0` , and remove the `enforce-semver2` option in your bootstrap command.

scbizu:
Cool @cbuto , it really do me a favor 🥳

PR #522, opened 2022-01-22
feat: deprecate enforcesemver2 config option

cbuto:
This is a follow up on #482 to fully "deprecate" the `--enforce-semver2` flag. It seems like some of the documentation work is capture in https://github.com/helm/chartmuseum/issues/485, which I can follow up on. 

Also, do we want a way to mark config options as deprecated and print warnings if they are set on startup?

### DIFF
--- a/pkg/chartmuseum/server.go
+++ b/pkg/chartmuseum/server.go
@@ -66,8 +66,7 @@ type (
 		CORSAllowOrigin        string
 		ReadTimeout            int
 		WriteTimeout           int
-		// EnforceSemver2 represents if the museum server always accept the Chart with [valid semantic version 2](https://semver.org/)
-		// More refers to : https://github.com/helm/chartmuseum/issues/320
+		// EnforceSemver was deprecated, see https://github.com/helm/chartmuseum/issues/485 for more info
 		EnforceSemver2 bool
 		CacheInterval  time.Duration
 		Host           string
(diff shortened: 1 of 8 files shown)


pkg/chartmuseum/server/multitenant/api.go
```diff
@@ -96,9 +95,20 @@ func (server *MultiTenantServer) deleteChartVersion(log cm_logger.LoggingFn, rep
...
+		// left the others fields to be default
+	})
+	if err != nil {
+		return filename, &HTTPError{http.StatusBadRequest, err.Error()}
+	}
+
+	filename, err = cm_repo.ChartPackageFilenameFromContent(content)
 	if err != nil {
-		// TODO: this is an error if invalid semver, making "EnforceSemver2" useless...
 		return filename, &HTTPError{http.StatusInternalServerError, err.Error()}
```
cbuto, line 112: While I get familiar with everything, should we return a `400` here and skip the call to `ChartVersionFromStorageObject` entirely?

jdolitsky (changes requested):
> do we want a way to mark config options as deprecated and print warnings
yes, that would be great. Thanks @cbuto!

pkg/chartmuseum/server/multitenant/api.go
cbuto: cool, fixed in 2f11e5029b735bc6a92a9fc566a732dfccfa9c01!

pkg/chartmuseum/server.go
```diff
@@ -137,9 +136,10 @@ func NewServer(options ServerOptions) (Server, error) {
 		UseStatefiles:          options.UseStatefiles,
 		AllowOverwrite:         options.AllowOverwrite,
 		AllowForceOverwrite:    options.AllowForceOverwrite,
-		EnforceSemver2:         options.EnforceSemver2,
-		Version:                options.Version,
-		CacheInterval:          options.CacheInterval,
+		// EnforceSemver was deprecated, see https://github.com/helm/chartmuseum/issues/485 for more info
+		EnforceSemver2: options.EnforceSemver2,
+		Version:        options.Version,
+		CacheInterval:  options.CacheInterval,
```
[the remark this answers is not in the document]
cbuto: done in 2f11e5029b735bc6a92a9fc566a732dfccfa9c01 👍

jdolitsky (approved):
👍

scbizu:
Cool @cbuto , it really do me a fever 🥳

scbizu:
Cool @cbuto , it really do me a favor 🥳

PR #533, opened 2022-01-28
feat: Add configuration deprecation warning logs

cbuto:
A follow up on https://github.com/helm/chartmuseum/pull/522 to print warnings if deprecated options are being set.

Admittedly, I haven't added unit tests around this yet as that required some refactors around how the configuration code works right now (a lot of the configuration relies on variables that aren't easily overridden). I can come back and handle that if this approach seems good!

It seemed the best way to approach this was to create the logger earlier on, though I didn't pull `Debug` and `LogJSON` out of the `ServerOptions` struct just in case for some reason these are being used by someone (seems unlikely).

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/chartmuseum/storage"
 	"helm.sh/chartmuseum/pkg/cache"
 	"helm.sh/chartmuseum/pkg/chartmuseum"
+	cm_logger "helm.sh/chartmuseum/pkg/chartmuseum/logger"
 	"helm.sh/chartmuseum/pkg/config"
 
 	"github.com/urfave/cli"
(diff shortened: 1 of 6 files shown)


cmd/chartmuseum/main.go
```diff
@@ -76,10 +88,8 @@ func cliHandler(c *cli.Context) {
 		ChartPostFormFieldName: conf.GetString("chartpostformfieldname"),
 		ProvPostFormFieldName:  conf.GetString("provpostformfieldname"),
 		ContextPath:            conf.GetString("contextpath"),
-		LogJSON:                conf.GetBool("logjson"),
```
jdolitsky, line 79: If LogJSON and Debug fields are no longer effective can you add a comment in ServerOptions prefixed with `// Deprecated: <reason>`?

jdolitsky (changes requested):

cmd/chartmuseum/main.go
cbuto: definitely, added in d84441b9718a08b412716104fb03fe9d71b8fb3a!

jdolitsky (approved):
excellent, thanks @cbuto !

PR #538, opened 2022-01-30
Generate SPDX SBOM at release time

### LINKED ISSUE #537: Add SBOM to release artifacts

puerco:
This PR adds the Kubernetes SBOM tool to the release workflow to generate a Software Bill of Materials.

There are two parts to achieve this, each split into a commit in the PR:

1. The actions workflow now uses the new [`bom-installer`](https://github.com/puerco/bom-installer) action to install bom into the runner.
2. The actual SBOM generation is done by a new script in `scripts/sbom.sh` that reads the artifacts and generates the SBOM including a) the source code and dependencies b) the container image c) the built tarballs.

/cc: @jdolitsky 

Fixes #537 

#### TODO: 

The current change only writes the SBOM. It should then be stored in its final location

#### Structure

This is a sample visualization of the generated SBOM:

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,8 @@ jobs:
         uses: sigstore/cosign-installer@1e95c1de343b5b0c23352d6417ee3e48d5bcd422
         with:
           cosign-release: 'v1.5.0'
+      - name: Install Kubernetes SBOM Tool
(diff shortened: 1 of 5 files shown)


jdolitsky (approved):
@puerco thank you! 🎉 

This is great. Made a few changes to integrate with the release pipeline (I think that is related to your TODO item)

PR #544, opened 2022-02-04
ci: move building the go binaries outside of the Docker image to speed up builds

### LINKED ISSUE #539: Review use of docker buildx in CI
The current build takes over 1 hour. Perhaps at the very least the buildx build and push steps can combined?

cbuto:
Related to #539. I tested this locally as much as possible using `act` with some of the build steps commented out.

This commit moves the building of the chartmuseum binaries out of the Docker image and uses `make build-cross` early on in the job to build all the binaries. Then the Docker image will copy in the proper binary based on the `TARGETARCH` build arg.

Building the images is now super quick, although some of that time was moved to the `make build-cross` step. However, now we only have to download go modules once instead of N number of times (where N = number of architectures).

I also combined the build and push steps since it seemed like an innocent change (like @jdolitsky mentioned in the issue) 🤞.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,9 +2,6 @@
 
 *
 
-# Bring in our source code and dependency listings
+# Bring in our binary
 
-!go.mod
-!go.sum
-!cmd/**
-!pkg/**
+!_dist/linux-*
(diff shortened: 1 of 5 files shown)


jdolitsky (approved):
Lets throw it against CI and see what happens here :)

`main` should produce canary image(s) so we can test them out

PR #549, opened 2022-02-09
refactor: use sync map instead for map string

nerdeveloper:
Signed-off-by: nerdeveloper <odirionye@gmail.com>

### DIFF
--- a/pkg/chartmuseum/server/multitenant/server.go
+++ b/pkg/chartmuseum/server/multitenant/server.go
@@ -49,7 +49,7 @@ type (
 		StorageBackend         cm_storage.Backend
 		TimestampTolerance     time.Duration
 		ExternalCacheStore     cache.Store
-		InternalCacheStore     map[string]*cacheEntry
+		InternalCacheStore     sync.Map
 		MaxStorageObjects      int
 		IndexLimit             int
 		AllowOverwrite         bool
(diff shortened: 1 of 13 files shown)


nerdeveloper:
Yes

scripts/test.sh
```diff
@@ -8,7 +8,7 @@ mkdir .cover/ .test/
 trap "rm -rf .test/" EXIT
 
 for pkg in `go list ./... | grep -v /vendor/`; do
-    go test -v -covermode=atomic \
+    go test -race -v -covermode=atomic \
```
cbuto, line 11: you'll need this set so the tests run!
```suggestion
    CGO_ENABLED=1 go test -race -v -covermode=atomic \
```

pkg/chartmuseum/server/multitenant/server.go
```diff
@@ -49,7 +49,7 @@ type (
 		StorageBackend         cm_storage.Backend
 		TimestampTolerance     time.Duration
 		ExternalCacheStore     cache.Store
-		InternalCacheStore     map[string]*cacheEntry
+		InternalCacheStore     sync.Map
```
scbizu, line 52: Can we wrap the `sync.Map` inside a specialized structure , for example `type memoryCacheStore struct { sync.Map }` maybe ,  and wrap the `Load` and `Store` func ?  Doing a type assertion every where  seems ugly though .

scbizu:
@nerdeveloper  And please fix your DCO verification ?

scbizu (changes requested):

cbuto:
@nerdeveloper is there anything we can help with here?

nerdeveloper:
Yes, @cbuto, I tried to refactor to sync map, but the race condition still exists. i am having a hard time debugging this

nerdeveloper:
I am going to close this PR, as @cbuto has picked it and finished it

Issue #447, opened 2021-04-11
/metrics is not protected with basic auth if enabled


agoloAbbas:
after adding the following options, /metrics still available to get accessed with no authentication 
```
--basic-auth-user=<user>
--basic-auth-pass=<pass>
```

shouldn't `/metrics` be protected too as it can contain sensitive data?

jdolitsky:
I'm not sure about that, since this is specifically for use by Prometheus, and I'm not sure how Prometheus would get these credentials..

Maybe the better option is to make these disabled by default (as they are in the Helm chart)

PR #554, opened 2022-02-17
refactor: remove disablemetrics for enablemetrics

nerdeveloper:
Signed-off-by: nerdeveloper <odirionye@gmail.com>

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -95,7 +95,7 @@ func cliHandler(c *cli.Context) {
 		UseStatefiles:          !conf.GetBool("disablestatefiles"),
 		AllowOverwrite:         conf.GetBool("allowoverwrite"),
 		AllowForceOverwrite:    !conf.GetBool("disableforceoverwrite"),
-		EnableMetrics:          !conf.GetBool("enablemetrics"),
+		EnableMetrics:          conf.GetBool("enablemetrics"),
 		AnonymousGet:           conf.GetBool("authanonymousget"),
 		GenIndex:               conf.GetBool("genindex"),
 		MaxStorageObjects:      conf.GetInt("maxstorageobjects"),
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -95,7 +95,7 @@ func cliHandler(c *cli.Context) {
 		UseStatefiles:          !conf.GetBool("disablestatefiles"),
 		AllowOverwrite:         conf.GetBool("allowoverwrite"),
 		AllowForceOverwrite:    !conf.GetBool("disableforceoverwrite"),
-		EnableMetrics:          !conf.GetBool("disablemetrics"),
+		EnableMetrics:          !conf.GetBool("enablemetrics"),
 		AnonymousGet:           conf.GetBool("authanonymousget"),
 		GenIndex:               conf.GetBool("genindex"),
 		MaxStorageObjects:      conf.GetInt("maxstorageobjects"),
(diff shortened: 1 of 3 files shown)


cmd/chartmuseum/main.go
```diff
@@ -95,7 +95,7 @@ func cliHandler(c *cli.Context) {
 		UseStatefiles:          !conf.GetBool("disablestatefiles"),
 		AllowOverwrite:         conf.GetBool("allowoverwrite"),
 		AllowForceOverwrite:    !conf.GetBool("disableforceoverwrite"),
-		EnableMetrics:          !conf.GetBool("disablemetrics"),
+		EnableMetrics:          !conf.GetBool("enablemetrics"),
```
scbizu, line 98: Should be `conf.GetBool("enablemetrics")` ?

scbizu (changes requested):

scbizu:
BTW , what issue the PR points to ?

nerdeveloper:
@scbizu the purpose of the PR is to disable metrics by default; this is the issue: https://github.com/helm/chartmuseum/issues/447

cmd/chartmuseum/main.go
nerdeveloper: i thought `conf.GetBool("enablemetrics")` will enable the metrics by default. we are trying to disable it

cbuto: You can use `conf.GetBool("enablemetrics")` and default the value to `false` to achieve the same thing, it just makes things a bit cleaner.

cbuto:
Just a note, we’ll also want to reflect this change in the [helm chart](https://github.com/chartmuseum/charts) and the [docs](https://github.com/chartmuseum/www) after it’s released. @nerdeveloper let us know if you are up for that!

cmd/chartmuseum/main.go
nerdeveloper: Gotcha!

nerdeveloper:
Sure, I am definitely up for that! I can see on the docs: https://chartmuseum.com/docs/#prometheus-metrics

Also, I will refactor it and wait till `v0.15.0` is out cc @jdolitsky

scbizu:
Agree with @cbuto's idea , and this issue should be related to the charts and doc PR .

pkg/config/vars.go
```diff
@@ -94,9 +94,19 @@ var configVars = map[string]configVar{
 		Default: false,
 		CLIFlag: cli.BoolFlag{
 			Name:   "disable-metrics",
-			Usage:  "disable Prometheus metrics",
+			Usage:  "(deprecated)disable Prometheus metrics",
```
jdolitsky, line 97: can we add a space between `(deprecated)` and `disable Prometheus metrics` ?

jdolitsky (changes requested):
Thanks for tackling this.

Can you also modify the README section to describe the new behavior?
https://github.com/helm/chartmuseum#prometheus-metrics

Maybe even mention the now deprecated old flag in versions `v0.14.0` and prior

pkg/config/vars.go
nerdeveloper: Sure, I will into this today.

nerdeveloper:
> Thanks for tackling this.
> 
> Can you also modify the README section to describe the new behavior? https://github.com/helm/chartmuseum#prometheus-metrics
> 
> Maybe even mention the now deprecated old flag in versions `v0.14.0` and prior

Definitely

nerdeveloper:
I have update chart museum Helm Chart for the Prometheus config here: https://github.com/chartmuseum/charts/pull/38

nerdeveloper:
OMG! Yes, it has been merged! I am so happy. thank you, team.

PR #565, opened 2022-03-16
Fix : Call Update Metrics when founding existing index-cache.yaml

jojolb:
Hi !

I have Chartmuseum deployed in a k8s cluster and connected to an Azure Blob Storage Backend.

When the remote index-cache.yaml is already present, metrics `charts_served_total` & `chart_versions_served_total` are not initialized, and the /metrics endpoint does not return those two metrics.

I suggest that if we found an index-cache.yaml when initializing cache entry, we force an "update metrics" on it.

Thank you !

### DIFF
--- a/pkg/chartmuseum/server/multitenant/cache.go
+++ b/pkg/chartmuseum/server/multitenant/cache.go
@@ -460,12 +460,15 @@ func (server *MultiTenantServer) newRepositoryIndex(log cm_logger.LoggingFn, rep
 		"repo", repo,
 	)
 
-	return &cm_repo.Index{
+	index := &cm_repo.Index{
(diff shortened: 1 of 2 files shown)


cbuto:
@jojolb thanks for the PR! it looks good but we'll want to check if we need any locking on the Index when calling UpdateMetrics. The locking mechanism is being reworked in https://github.com/helm/chartmuseum/pull/556, after that goes through we'll want to rebase and test these changes!

cbuto:
@jojolb now that https://github.com/helm/chartmuseum/pull/556 has been merged, can you rebase with main so we can run the tests? thanks!

jdolitsky:
closing as outdated

Issue #323, opened 2020-04-07
no Read-After-Write consistency


samrees:
A client we use to access chartmuseum uploads charts then immediately tries to fetch them down. https://github.com/cathive/concourse-chartmuseum-resource/blob/7dab3bf0168f32dbd6632daa8571a4958c7e1145/out.ts#L267

Recently we've been seeing a failure where that immediate GET returns 
{"error":"No chart version found for xxxxx-0.59.3"}

Since the backing S3 storage is read-after-write consistent, we expected ChartMuseum to be the same, but it doesnt appear that it is. 

Maybe this shouldnt be expected to be supported, or if it supported it will need to be supported through a different PUT call that blocks until all chartmuseums are capable of GETing it, but thought to open this issue in case others are seeing this.

samrees:
Maybe related to this: https://github.com/helm/chartmuseum/issues/175

Issue #453, opened 2021-04-27
 index.yaml not always regenerated when pushing helm charts via API - 404s


peppe77:
K8S: 1.18
Chartmuseum: v0.13.1 (no cache-interval configured)

```env:
        - name: DISABLE_METRICS
          value: "false"
        - name: DISABLE_API
          value: "false"
        - name: ALLOW_OVERWRITE
          value: "true"
        - name: CACHE
          value: redis
```

```
GET /ccdp/stable/api/charts/ccdp-om-inbound-service/release-1.61.0-dev699051 
HTTP/1.1 404 Not Found
Date: Tue, 27 Apr 2021 22:21:32 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 57
Connection: keep-alive
X-Request-Id: dc61e2062300687cbae84e6fc9f744a6
Strict-Transport-Security: max-age=15724800; includeSubDomains
{"error":"improper constraint: release-1.61.0-dev699051"}
```

Though when we checked in the S3 bucket under /ccdp/stable, we do find the helm chart [ccdp-om-inbound-service-release-1.61.0-dev699051]. Upload always seems to work though not always index.yaml gets regenerated. Long story short, if a 404 is returned then we do not deploy (not in automated manner).

jdolitsky:
Initial reaction is that this is related to the version of your chart not being valid semver2

```
{"error":"improper constraint: release-1.61.0-dev699051"}
```

This error is triggered by this code: https://github.com/Masterminds/semver/blob/49c09bfed6adcffa16482ddc5e5588cffff9883a/constraints.go#L32

which is referenced all over in helm/helm

modify the version from `release-1.61.0-dev699051` to `1.61.0-dev699051` (or something else starting with valid semver) and I bet this issue will go away

peppe77:
@jdolitsky it was a bad example. The logic looks for 1.61.0-dev699051, 1.61.0, 1.161.0-dev and also release-1.61.0-dev699051(this is only case a 404 should be returned, right?). So, will reproduce and then do a get following this pattern [1.61.0-dev<string>, 1.61.0, 1.161.0-dev ] and report back.

peppe77:
@jdolitsky here it is
```GET https://helm.infrastructure.volvo.care/api/ccdp/stable/charts/ccdp-best-match/1.62.0-dev699615
404
GET /api/ccdp/stable/charts/ccdp-best-match/1.62.0-dev699615 HTTP/1.1
[...]

PR #574, opened 2022-04-12
feat: Add read-after-write consistency configuration option

cbuto:
Closes https://github.com/helm/chartmuseum/issues/323.

This PR adds a configuration option `--read-after-write-consistency` to allow users to choose whether ChartMuseum processes uploads/deletion events synchronously or asynchronously. This also might help with situations like https://github.com/helm/chartmuseum/issues/453, where users expect charts to be immediately available after uploading (although there could likely still be a bug causing that issue, I couldn't reproduce when using `main`). 

let me know what you think about supporting this overall!


pkg/chartmuseum/server/multitenant/api.go
```diff
@@ -281,6 +281,8 @@ func (server *MultiTenantServer) PutWithLimit(ctx *gin.Context, log cm_logger.Lo
 	if err = server.StorageBackend.PutObject(pathutil.Join(repo, filename), content); err != nil {
 		return fmt.Errorf("PutWithLimit: put new chart: %w", err)
 	}
+
+	// TODO: I think emitting an event here is fine if read-after-write consistency is set
```
cbuto, line 285: looking for some feedback here, I don't think we need this to happen synchronously but let me know what you think!

cbuto:
found an issue while testing, will reopen soon!

Issue #419, opened 2021-02-20
How to upload meta file artifacthub-repo.yml?


acim:
This is required by ArtifactHub to verify ownership and should be visible at the similar URL as index.yaml, but I see no way how to add it to chartmuseum.

scbizu:
AFAIK we do not registered in ArtifactHub yet . And It seems that ArtifactHub need the metadata file to claim the ownership ?

acim:
Well, thisi optional, but if you want to get badge "Verified publisher", you need to have this file and repoID from the ArtifcacttHub inside. I solved this at the moment by nginx ingress rewrite rule, but it would be nice if you support it.

scbizu:
Thanks for your advice .

/cc @jdolitsky

jdolitsky:
Seems like we can start with a flag like `--artifact-hub-repo-id=xxxxx`, and just populate the `repositoryID` field here: https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-repo.yml#L12

@acim - do you have any need for the `owners` / `ignore` fields?

acim:
At this moment this would be enough. Thank you.

PR #579, opened 2022-04-28
feat: Add support for hosting an artifacthub-repo.yml file

cbuto:
Fixes https://github.com/helm/chartmuseum/issues/419.

This is an initial take at hosting an `artifacthub-repo.yml` from ChartMuseum.

Currently, it will only support setting the `repositoryID` in the `artifacthub-repo.yml` file, but adding more fields in the future shouldn't be too difficult.

Let me know what you think, especially the CLI flag implementation for multitenancy.

### DIFF
--- a/cmd/chartmuseum/main.go
+++ b/cmd/chartmuseum/main.go
@@ -116,6 +116,7 @@ func cliHandler(c *cli.Context) {
 		Host:                   conf.GetString("listen.host"),
 		PerChartLimit:          conf.GetInt("per-chart-limit"),
 		WebTemplatePath:        conf.GetString("web-template-path"),
+		ArtifactHubRepoID:      conf.GetString("artifact-hub-repo-id"),
 	}
 
 	server, err := newServer(options)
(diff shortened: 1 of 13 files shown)


cmd/chartmuseum/main.go
```diff
@@ -116,6 +117,7 @@ func cliHandler(c *cli.Context) {
 		Host:                   conf.GetString("listen.host"),
 		PerChartLimit:          conf.GetInt("per-chart-limit"),
 		WebTemplatePath:        conf.GetString("web-template-path"),
+		ArtifactHubRepoID:      conf.GetStringMapString("artifact-hub-repo-id"),
```
scbizu, line 120: will it be nil if no flag set ? Will be panic if no `artifact-hub-repo-id` flag set but still visit the `/artifact-repo.yaml` ?

scbizu:
For me , I can accept the flag set mechanism for multi-tenancy but it maybe hard to set those flag if users have a large set of tenancy , any thoughts about the stdin (i.e.: json file stdin) ?

scbizu:
>  Maybe supporting a combination of the two could be implemented

Yeah , I mean that .

jdolitsky (approved):
looks great!

scbizu (approved):

PR #601, opened 2022-07-05
build: bump cosign to v1.9.0

cbuto:


.github/workflows/build.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       - name: Install sigstore cosign
         uses: sigstore/cosign-installer@1e95c1de343b5b0c23352d6417ee3e48d5bcd422
```
jdolitsky, line 76: can we use `@48866aa521d8bf870604709cd43ec2f602d03ff2` or `@main`?

.github/workflows/build.yml
```diff
@@ -75,7 +75,7 @@ jobs:
...
         with:
-          cosign-release: 'v1.5.0'
+          cosign-release: 'v1.8.0'
```
jdolitsky, line 78: v1.9.0 is latest

jdolitsky (changes requested):

.github/workflows/build.yml
cbuto: done!

jdolitsky (approved):

PR #615, opened 2022-09-06
[WIP] feat: Use Redis transactions for caching

cbuto:
This should fix the problems described in https://github.com/helm/chartmuseum/issues/453

To avoid data consistency issues when multiple goroutines process cacheEntries, we can use Redis transactions to retry
the transaction logic if the cache entry changed during the transaction.

I'm open to suggestions about the current implementation. I tried to abstract away the transaction handling behind the store interface. I do think there's more changes that could make it cleaner by adopting a repository pattern and also moving the in memory cache implementation behind the store interface.


jdolitsky (approved):
overall lgtm!

scbizu:
LGTM , but I am interest with the in-memory implementation with transaction , any ideas yet ?

cbuto:
I did some testing with the locust test suite and found something I'll want to handle (maybe in a follow up PR). 

This [transaction](https://github.com/helm/chartmuseum/pull/615/files#diff-b650a52a32945b29a148c4710ad9c7feac77eb281347691bae543c77d1afc6b2R89) is really large and can run for a long time because we update every chart in the index every time it runs. That leads to really poor performance with a large number of charts and with this change could block subsequent requests.

I think the easiest thing to do would be to split the transaction up if possible, I'll look into that.

cbuto:
I'm going to run some more load tests to see how this affects performance as compared to main

cbuto:
here are the results of some minimal stress tests I ran locally:

Test 1:

* branch: fix/redis-consistency
* Local cache, AWS s3 storage
* 50 users/10 spawn rate
* 3985 charts uploaded - all charts found in index
test1-local-cache-3985-charts-fix:redis


Test 2:

* branch: fix/redis-consistency
* Redis cache (local docker container), AWS s3 storage
* 50 users/10 spawn rate
* 1418 charts uploaded - all charts found in index
test2-redis-1418-charts-fix:redis


Test 3:
* branch: main
* Local cache, AWS s3 storage
* 50 users/10 spawn rate
* 4067 charts uploaded - all charts found in index

test3-local

Test 4:

* branch: main
* Redis cache, AWS s3 storage
* 50 users/10 spawn rate
* 1579 charts uploaded - some charts missing from the index

test4-redis

jdolitsky:
closing as outdated

PR #642, opened 2022-11-11
support for delete all version of a specific chart

carlclone:


kladiv:
+1 really required feature

PR #667, opened 2023-02-20
fix: update URLs in k8s mirror script

Syoc:
Pointing URLs in script to the new ones ref. https://helm.sh/blog/new-location-stable-incubator-charts/


scbizu:
Thanks

PR #703, opened 2023-06-30
Prep for v0.16.1 release

jdolitsky:


README.md
```diff
@@ -95,14 +95,8 @@ helm install chartmuseum/mychart --generate-name
 ## How to Run
 ### CLI
 #### Installation
-Install binary using [GoFish](https://gofi.sh/):
-```
-gofish install chartmuseum
-==> Installing chartmuseum...
-🐠  chartmuseum 0.16.0: installed in 95.431145ms
-```
```
jdolitsky, line 103: gofish has been archived so we can make updates there anymore

Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION ?= 0.16.0
+VERSION ?= 0.16.1
 REVISION := $(shell git rev-parse --short HEAD;)
```
jdolitsky, line 2: we should probably figure out a way to get rid of these versions everywhere so we can release more easily

cbuto: definitely, should be able to get the tag from git/maybe introduce goreleaser

cbuto (approved):

PR #717, opened 2023-08-28
chore: add golint for checking linting

nerdeveloper:
add linting to CI

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -19,6 +19,12 @@ jobs:
         run: make bootstrap
       - name: run unit tests
         run: make test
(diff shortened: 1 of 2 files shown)


.github/workflows/build-pr.yml
```diff
@@ -19,6 +19,12 @@ jobs:
         run: make bootstrap
       - name: run unit tests
         run: make test
+      - name: lint
```
cbuto, line 22: wdyt about running the linter before running unit tests?

nerdeveloper: my bad! must have missed that, I will fix that

Issue #715, opened 2023-08-22
about raw data api


allenhaozi:
For our scenario, we'll add a static/deploy.j2 to helm chart to render the helm charts information

So we need to get this file through the api,

However, the current api does not support it, as follows:

```go

chartManipulationRoutes := []*cm_router.Route{
		{Method: "GET", Path: "/api/:repo/charts", Handler: s.getAllChartsRequestHandler, Action: cm_auth.PullAction},
		{Method: "HEAD", Path: "/api/:repo/charts/:name", Handler: s.headChartRequestHandler, Action: cm_auth.PullAction},
		{Method: "GET", Path: "/api/:repo/charts/:name", Handler: s.getChartRequestHandler, Action: cm_auth.PullAction},
		{Method: "HEAD", Path: "/api/:repo/charts/:name/:version", Handler: s.headChartVersionRequestHandler, Action: cm_auth.PullAction},
		{Method: "GET", Path: "/api/:repo/charts/:name/:version", Handler: s.getChartVersionRequestHandler, Action: cm_auth.PullAction},
		{Method: "GET", Path: "/api/:repo/charts/:name/:version/templates", Handler: s.getStorageObjectTemplateRequestHandler, Action: cm_auth.PullAction},
		{Method: "GET", Path: "/api/:repo/charts/:name/:version/values", Handler: s.getStorageObjectValuesRequestHandler, Action: cm_auth.PullAction},
		{Method: "POST", Path: "/api/:repo/charts", Handler: s.postRequestHandler, Action: cm_auth.PushAction},
		{Method: "POST", Path: "/api/:repo/prov", Handler: s.postProvenanceFileRequestHandler, Action: cm_auth.PushAction},
	}

```
In the helm charts definition, the raw field can be as follows:
```go
type Chart struct {
	// Raw contains the raw contents of the files originally contained in the chart archive.
	//
	// This should not be used except in special cases like `helm show values`,
	// where we want to display the raw values, comments and all.
	Raw []*File `json:"-"`
	// Metadata is the contents of the Chartfile.
	Metadata *Metadata `json:"metadata"`
	// Lock is the contents of Chart.lock.
	Lock *Lock `json:"lock"`
	// Templates for this chart.
	Templates []*File `json:"templates"`
	// Values are default config for this chart.
	Values map[string]interface{} `json:"values"`
	// Schema is an optional JSON schema for imposing structure on Values
	Schema []byte `json:"schema"`
[...]

PR #719, opened 2023-09-08
support raw api (#715)

allenhaozi:
suport raw api

Hi @cbuto @scbizu, as for this raw api, submit a pr version, 

please review it. If you have any questions, please feedback to me. Thank you

#715


pkg/chartmuseum/server/multitenant/handlers.go
```diff
@@ -186,6 +183,34 @@ func (server *MultiTenantServer) getStorageObjectTemplateRequestHandler(c *gin.C
...
+		c.JSON(http.StatusNotFound, gin.H{"error": err.Message})
+		return
+	}
+
+	storageObject, err := server.getStorageObject(log, repo, fileName)
+	if err != nil {
+		c.JSON(err.Status, gin.H{"error": err.Message})
+		return
+	}
+	chrt, err1 := loader.LoadArchive(bytes.NewReader(storageObject.Content))
```
scbizu, line 204: Can we just put this in `pkg/repo/chart.go`, and provide a function like `GetChartRawdate()` ?

scbizu (changes requested):

scbizu:
And also , could you add some testing code to prove the feature is worked as expected ?

pkg/chartmuseum/server/multitenant/handlers.go
allenhaozi: thank you for your review
I tried to wrap a function like you said, but I got an error `import cycle not allowed`
That would require a bigger change, what's your opinion? whether this larger change is needed

PR #721, opened 2023-09-17
pkg/repo,pkg/chartmuseum/server: per-limit-chart chart expiration removes prefix mathching for chartpath preparing

### LINKED ISSUE #714: PER_CHART_LIMIT name prefix filtering deletes unrelated charts
We've noticed a strange issue where charts are - for some reason we can't determine - disappearing from our S3 bucket.

Versions:
*  0.16.0 / Helm chart  3.10.1
* K8s 1.27.x

We're using "<service_name>" and "<service_name>-<branch>" as the name of the charts we upload, to distinguish production from test charts, and to pin the number of charts within each we have `PER_CHART_LIMIT` set to 10. We know that this mechanism is working fine, and have watched the oldest chart of 10 deleted before the newest one is stored for _that particular_ service/branch. All good.

What is different is that when we check, say "clients-dev", that the folder is unexpectedly empty, and this seems to be completely independent of the above mechanism, e.g. I doubt there have ever been 10 charts in that particular folder.

So are there are any _other_ expiry mechanisms working behind the scenes? Eyeballing the source suggests not, but I can't be 100% sure.

This clearly isn't related to "age" of chart per se, as the very first / oldest charts we added are still present. The deleted charts are likely much younger.

We're looking to see if we can get an audit log enabled to track the deletes. Before you ask, I can pretty much 100% rule out manual deleting...

I set `DEBUG` to true, but that doesn't give us any additional visibility of deletes.

How common is this? Not common, no. The two occasions this was noticed were 21 days apart, but each time several charts were found to be missing. I can't be any more specific about timings without the audit log.

---

Addendum: we don't have any lifecycle rules on the S3 bucket

scbizu:
Fix #714


scbizu:
@cbuto  PTAL

cbuto (approved):
lgtm 🚀

PR #748, opened 2024-01-31
Update index.go

wutongjie23hao:
the chartversions of index.yaml should be equal to the length of chartversion objects in storage backend. if not equal, we should regenerate the index.yaml.


scbizu:
I prefer to not accept this PR , since we already have internal-cache option and always-regenerate-chart-index option on the other hand , this kind of feature seems duplicate . What do you think ? @cbuto

wutongjie23hao:
> I prefer to not accept this PR , since we already have internal-cache option and always-regenerate-chart-index option on the other hand , this kind of feature seems duplicate . What do you think ? @cbuto

always regenerate index only when the depth set a certain value. This is not valid when we set  flag '--depth-dynamic'.

PR #769, opened 2024-05-09
Bump aws-sdk-go to 1.48.7 to support EKS Pod Identity

notwedtm:
When attempting to use a credential with EKS Pod Identities the following error is emitted:

```
2024/05/09 04:04:39 Ignoring, HTTP credential provider invalid endpoint host, "169.254.170.23", only loopback hosts are allowed. <nil>
```

This is resolved by upgrading the SDK to `v1.48.7`.


scbizu:
The [AWS Doc](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-minimum-sdk.html) says v1.47.11 has the EKS Pod Identity , have you tried our `canary` image ?

PR #774, opened 2024-05-30
Bump version to prep for 0.16.2

cbuto:
bumps the version to prep for a patch release

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,7 +472,7 @@ docker run --rm -it \
   -e STORAGE=local \
   -e STORAGE_LOCAL_ROOTDIR=/charts \
   -v $(pwd)/charts:/charts \
-  ghcr.io/helm/chartmuseum:canary
+  ghcr.io/helm/chartmuseum:v0.16.2
 ```
 
 Example usage (S3):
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -472,7 +472,7 @@ docker run --rm -it \
   -e STORAGE=local \
   -e STORAGE_LOCAL_ROOTDIR=/charts \
   -v $(pwd)/charts:/charts \
-  ghcr.io/helm/chartmuseum:v0.16.1
+  ghcr.io/helm/chartmuseum:canary
```
scbizu, line 475: can we add some explanation here that the canary is our main version after the action is done , most people asked it in the issue list . I'd prefer keep the tag here. And also we should update www .

scbizu (changes requested):

README.md
cbuto: I reverted to use the released version. eventually we'll want to remove/automate this to make releases easier

jdolitsky (approved):

scbizu (approved):

scbizu:
LGTM 🎉

PR #791, opened 2024-11-06
feat: allow HEAD on :repo/charts/:filename endpoint

dmytroba-salt:


dmytroba-salt:
@cbuto please consider taking a look at this

Issue #1068, opened 2025-03-16
v0.16.3 Release


scbizu:
Prepare for releasing the v0.16.3 , with normal security patch updates and 1 bugfix 

See https://github.com/helm/chartmuseum/milestone/8

@jdolitsky @cbuto Can you help with it ?

Ravikirandg29:
@scbizu Could you please provide ETA for this release?

cbuto:
https://github.com/helm/chartmuseum/releases/tag/v0.16.3

PR #1075, opened 2025-04-14
Bump version to prep for 0.16.3

cbuto:
Bumps version to prep for 0.16.3

part of https://github.com/helm/chartmuseum/issues/1068

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-VERSION ?= 0.16.2
+# TODO(cbuto): Pull from git tag to avoid needing to bump this prior to releasing
+VERSION ?= 0.16.3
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,4 +1,5 @@
-VERSION ?= 0.16.2
+# TODO(cbuto): Pull from git tag to avoid needing to bump this prior to releasing
```
scbizu, line 1: Since we run this inside Action , I think this will be helpful . May be `$(shell git describe --tags --abbrev=0)` ? And then , we should run the action when every new tag pushed ? 🤔

scbizu (approved):

scbizu:
overall , LGTM .

PR #1081, opened 2025-05-18
fix: add trailing slash to repo path when listing objects

nerdeveloper:
This pull request includes a minor update to the `fetchChartsInStorage` function in the `pkg/chartmuseum/server/multitenant/cache.go` file. The change ensures that the `repo` parameter is joined with an empty string using `pathutil.Join` when calling the `ListObjects` method on `StorageBackend`.

To make it super clear:

```javascript
Before Fix:
└── bucket
    ├── abc-def-1.1.1.tgz    ❌ (ChartMuseum wrongly included these)
    ├── abc-def-1.1.24.tgz   ❌ (because they start with "abc")
    └── abc/
        └── xyz-1.1.1.tgz    ✅ (should only include this)

After Fix:
└── bucket
    ├── abc-def-1.1.1.tgz    ✅ (ChartMuseum correctly ignores these)
    ├── abc-def-1.1.24.tgz   ✅ (because they're not in the abc/ folder)
    └── abc/
        └── xyz-1.1.1.tgz    ✅ (only includes this, as it should!)
```


nerdeveloper:
i messed up my commits trying to verify commits

PR #1082, opened 2025-05-18
fix: ensure proper prefix for tenant repositories in storage operations

### LINKED ISSUE #794: Unable to refresh tenant index and get new chart versions
Hello,
I'm using multitenancy and alibaba OSS bucket as backend. My CM is not able to refresh the index of a tenant before a chart name's prefix at the root level is the same as the name of the tenant.

For example, in the bucket i have
```
/abc-def-1.1.1.tgz
...
/abc-def-1.1.24.tgz
/abc/xyz-1.1.1.tgz
```
If I uploaded a new chart
```
/abc/xyz-1.1.2.tgz
```
then my helm client can't see this new version because CM keeps failing to refresh `abc` index. 
```
{"L":"INFO","T":"2024-12-02T07:51:12.661Z","M":"Rebuilding index for tenant","repo":"abc"}
...
{"L":"DEBUG","T":"2024-12-02T07:51:12.661Z","M":"Entry found in cache store","repo":"abc"}
{"L":"DEBUG","T":"2024-12-02T07:51:12.661Z","M":"Fetching chart list from storage","repo":"abc"}
{"L":"DEBUG","T":"2024-12-02T07:51:12.676Z","M":"Change detected between cache and storage","repo":"abc"}
{"L":"DEBUG","T":"2024-12-02T07:51:12.676Z","M":"Regenerating index.yaml","repo":"abc"}
{"L":"DEBUG","T":"2024-12-02T07:51:12.676Z","M":"Loading charts packages from storage (this could take awhile)","repo":"abc","total":24}
{"L":"ERROR","T":"2024-12-02T07:51:12.680Z","M":"oss: service returned error: StatusCode=404, ErrorCode=NoSuchKey, ErrorMessage=\"The specified key does not exist.\", RequestId=674D66F04E22F83137933220","repo":"abc"}
```

And i think it's because of this line https://github.com/helm/chartmuseum/blob/8795e9930ad1f88f8c20426368327b0e500adcda/pkg/chartmuseum/server/multitenant/cache.go#L201

Is there suppose to be a `/` at the end? Not sure how it would affect the other backend suppliers

Thanks in advance!

nerdeveloper:
This PR fixes a critical tenant isolation bug in ChartMuseum:

Before the fix:

- When listing charts for tenant "abc", ChartMuseum incorrectly included any chart that started with "abc" (like "abc-def-1.1.1.tgz") even if it was in the root directory
- This was a security issue as tenants could potentially see charts they shouldn't have access to

After the fix:

- ChartMuseum now properly isolates tenant charts by only including charts that are actually inside the tenant's directory
- For example, for tenant "abc", only charts inside the "abc/" directory are included
- Charts in the root that just happen to start with "abc" are correctly excluded

To make it super clear:

```
Before Fix:
└── bucket
    ├── abc-def-1.1.1.tgz    ❌ (ChartMuseum wrongly included these)
    ├── abc-def-1.1.24.tgz   ❌ (because they start with "abc")
    └── abc/
        └── xyz-1.1.1.tgz    ✅ (should only include this)

After Fix:
└── bucket
    ├── abc-def-1.1.1.tgz    ✅ (ChartMuseum correctly ignores these)
    ├── abc-def-1.1.24.tgz   ✅ (because they're not in the abc/ folder)
    └── abc/
        └── xyz-1.1.1.tgz    ✅ (only includes this, as it should!)
```

### DIFF
--- a/pkg/chartmuseum/server/multitenant/api.go
+++ b/pkg/chartmuseum/server/multitenant/api.go
@@ -197,7 +197,12 @@ func (server *MultiTenantServer) uploadProvenanceFile(log cm_logger.LoggingFn, r
 
 func (server *MultiTenantServer) checkStorageLimit(repo string, filename string, force bool) (bool, error) {
 	if server.MaxStorageObjects > 0 {
-		allObjects, err := server.StorageBackend.ListObjects(repo)
+		// Ensure proper prefix for tenant repos by adding trailing slash
+		prefix := repo
+		if repo != "" {
+			prefix = repo + "/"
+		}
+		allObjects, err := server.StorageBackend.ListObjects(prefix)
 		if err != nil {
 			return false, err
 		}
@@ -246,7 +251,12 @@ func (server *MultiTenantServer) PutWithLimit(ctx *gin.Context, log cm_logger.Lo
 	defer server.ChartLimits.Unlock()
 	// clean the oldest chart(both index and storage)
 	// storage cache first
-	objs, err := server.StorageBackend.ListObjects(repo)
+	// Ensure proper prefix for tenant repos by adding trailing slash
+	prefix := repo
+	if repo != "" {
+		prefix = repo + "/"
+	}
+	objs, err := server.StorageBackend.ListObjects(prefix)
 	if err != nil {
 		return err
 	}
--- a/pkg/chartmuseum/server/multitenant/cache.go
+++ b/pkg/chartmuseum/server/multitenant/cache.go
@@ -198,7 +198,12 @@ func (server *MultiTenantServer) fetchChartsInStorage(log cm_logger.LoggingFn, r
 	log(cm_logger.DebugLevel, "Fetching chart list from storage",
 		"repo", repo,
 	)
-	allObjects, err := server.StorageBackend.ListObjects(repo)
+	// Ensure proper prefix for tenant repos by adding trailing slash
+	prefix := repo
+	if repo != "" {
+		prefix = repo + "/"
+	}
+	allObjects, err := server.StorageBackend.ListObjects(prefix)
 	if err != nil {
 		return []cm_storage.Object{}, err
 	}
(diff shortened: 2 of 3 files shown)


nerdeveloper:
I  am using the local‐filesystem backend (not a mock) inside the tests and added the behaviour change you mentioned (prefix isolation for multi-tenant depth). @scbizu, let me know

scbizu:
Hi @nerdeveloper , I am looking back this issue , and find that the unit testing code maybe unnecessary for this case , because our test code using local as backend provider . And the local storage impl [is correct](https://github.com/chartmuseum/storage/blob/main/local.go#L45) for this case . To keep the behavior is the similar from every backend provider , I prefer to change the storage package and not the CM end , what do you think ?

manysuq:
Took a proper look at this one since it's been sitting for a while. The bug is real and the fix is correct, I verified it against the storage backends:

- On S3/GCS-style backends, `ListObjects("abc")` does a plain prefix match, so it returns both `abc/x.tgz` and `abc-def.tgz`. `removePrefixFromObjectPath` then leaves `abc-def.tgz` untouched (the `abc/` replacement never matches), and `objectPathIsInvalid` doesn't filter it (no slash, not empty). So a chart sitting in the root leaks into tenant `abc`'s index. Adding the trailing slash fixes exactly this.
- On the local filesystem backend `ListObjects` reads the directory directly, so the change is neutral there.
- The depth-0 handling is good: keeping `""` when `repo` is empty avoids passing `"/"` as an S3 prefix, which would have been a regression.

I also checked that these three call sites (`cache.go` `fetchChartsInStorage`, `api.go` `checkStorageLimit`, `api.go` `PutWithLimit`) are still the only non-test `ListObjects` calls in the repo on current main, so this covers the full surface.

Two small things:

1. The prefix-building block is duplicated three times. A tiny helper like `tenantListPrefix(repo)` would keep it in one place.
2. Did you consider fixing this inside chartmuseum/storage instead, so every backend gets consistent semantics? If the reason is that it changes the public API of all ten backends, that's a perfectly fine reason to do it at this level instead, just worth stating in the PR description for future readers.

I only traced S3, GCS and local in detail; the remaining backends look like they follow the same plain-prefix pattern, but worth a glance at etcd since its listing semantics differ from the object stores.

The test suite is a nice addition. Hope this helps get it over the line, it looks mergeable to me.

scbizu:
@manysuq thanks for the testing , I will handle the merge and the releasing this weekend ~

